PR #15, opened 2015-01-10
update purescript-foldable-traversable-0.2.1

philopon:
Hi!

Purescript-foldable-traversable-0.2.1 has scanl, so it conflict with Data.List.Lazy.
I removed scanl from Data.List.Lazy to fix it.

Regards.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "purescript-arrays"               : "~0.3.0",
     "purescript-maybe"                : "~0.2.1",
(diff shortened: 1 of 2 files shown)


paf31:
Thanks!

garyb:
Unfortunately we should have updated the dependencies for `purescript-lazy` and bumped that at the same time before releasing this as `v0.4.0` - see https://github.com/purescript/purescript-foldable-traversable/pull/15#issuecomment-67421333

I guess we should re-release this as `v0.5.0` once lazy is bumped also.

PR #22, opened 2015-05-08
Make Apply consistent with Bind

puffnfresh:


paf31:
I'm going to only bump the minor version, since it's a fix, even though it changes behavior, because any code using this was probably broken anyway.

PR #23, opened 2015-05-08
Fix typo in ListT Apply

puffnfresh:


paf31:
Note to self: port this to `purescript-transformers` later.

PR #29, opened 2015-06-11
Update test-src to 0.7

sharkdp:
Fix the indentation and use `toList` for the `Arbitrary (List a)` instance.


paf31:
:+1: Thanks!

PR #30, opened 2015-06-14
Fix inconsistent behavior of range function

sharkdp:
The following behavior of the range function seems inconsistent to me (it *is* inconsistent with `Data.Array.range`)
```purescript
0 .. 2 == 0 : 1 : 2 : Nil
0 .. 1 == 0 : 1 : Nil
0 .. 0 == Nil -- would expect this to be 0 : Nil
```
The appended fix modifies the behavior of `range` to return `singleton 0` in the last case.


garyb:
:+1: thanks!

PR #31, opened 2015-06-14
Fix updateAt and alterAt to get tests running

sharkdp:
Fix two small mistakes in Data.List.Lazy

I'm not sure if I should include the changes in the docs.. but this is what psc-docs generates at the moment.


paf31:
Oops, thanks for this.

PR #32, opened 2015-06-16
Data.ZipList

paf31:
@garyb Could you please review? I thought this might be useful.


garyb:
:+1:

Should we stick it under `Data.List` with it being a newtype, rather than actually a different data structure?

paf31:
I think it's best in its own module personally, but I don't feel too strongly. We do follow this "pattern" elsewhere though: `Par`, `Parallel`, `Additive` etc.

garyb:
Sorry, yeah, I meant in the sense of `Data.List.ZipList`.

paf31:
Oh right, yeah. Good idea.

PR #35, opened 2015-07-26
Complexity, Generalizations of Monadic Functions, and Code Reuse

DavidHarrison:
- Add comments on the complexity of more functions
- Reduce the complexity of `findLastIndex`
- Generalize `replicateM` and `filterM` to `Applicative`
- Reuse existing functions where reasonable in definitions

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -244,7 +244,9 @@ head (Cons x _) = Just x
 -- |
 -- | Running time: `O(n)`.
 last :: forall a. List a -> Maybe a
-last = head <<< reverse
+last Nil          = Nothing
+last (Cons x Nil) = Just x
+last (Cons x xs)  = last xs
 
 -- | Get all but the first element of a list, or `Nothing` if the list is empty.
 -- |
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -399,7 +399,7 @@ filter p = foldMap (\x -> if p x then singleton x else Nil)
 -- | powerSet = filterM (const [true, false])
 -- | ```
 filterM :: forall a m. (Applicative m) => (a -> m Boolean) -> List a -> m (List a)
-filterM _ Nil = return Nil
+filterM _ Nil = pure Nil
 filterM p (Cons x xs) = consIf <$> p x <*> filterM p xs
   where consIf b xs' = if b then Cons x xs' else xs'
 
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -472,7 +472,7 @@ slice start end xs = take (end - start) (drop start xs)
 
 -- | Take the specified number of elements from the front of a list.
 -- |
--- | Running time: `O(2n)` where `n` is the number of elements to take.
+-- | Running time: `O(n)` where `n` is the number of elements to take.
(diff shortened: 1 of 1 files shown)


src/Data/List.purs
```diff
@@ -399,27 +398,17 @@ filter p = go Nil
...
 -- | Running time: `O(n)`
 mapMaybe :: forall a b. (a -> Maybe b) -> List a -> List b
-mapMaybe f = go Nil
-  where
-  go acc Nil = reverse acc
-  go acc (Cons x xs) =
-    case f x of
-      Nothing -> go acc xs
-      Just y -> go (Cons y acc) xs
+mapMaybe f = foldMap (maybe Nil singleton <<< f)
```
paf31: Is this definitely still O(n)? It only is if the singletons are concatenated to the left of the result.

src/Data/List.purs
```diff
@@ -476,12 +465,14 @@ sortBy cmp = mergeAll <<< sequences
...
 -- | Extract a sublist by a start and end index.
+-- |
+-- | Running time: `O(n + m)` where `n` and `m` are the start and end of the range.
 slice :: forall a. Int -> Int -> List a -> List a
 slice start end xs = take (end - start) (drop start xs)
 
 -- | Take the specified number of elements from the front of a list.
 -- |
--- | Running time: `O(n)` where `n` is the number of elements to take.
+-- | Running time: `O(2n)` where `n` is the number of elements to take.
```
paf31: What's the benefit of `2n` over `n` here?

paf31:
Looks great! I added a couple of comments.

src/Data/List.purs
DavidHarrison: I believe it should be the same as the above functions using `foldMap`, since `foldMap` is right fold, and concatenation to the singleton should be constant time. It does feel like a fair amount of wrapping/unwrapping, perhaps it would be better to define `mapMaybe` (and use it for filter as well)?

src/Data/List.purs
DavidHarrison: None that I'm aware of, I'll change it.

src/Data/List.purs
```diff
@@ -239,9 +244,7 @@ head (Cons x _) = Just x
 -- |
 -- | Running time: `O(n)`.
 last :: forall a. List a -> Maybe a
-last (Cons x Nil) = Just x
-last (Cons _ xs)  = last xs
-last _            = Nothing
+last = head <<< reverse
```
garyb: I'm pretty sure this will be a lot slower than the original definition - reconstructing the entire list and taking an element vs reading it once and taking the last element?

DavidHarrison: I hadn't been thinking of that. I'll try using a fold.

garyb: Currently TCO will optimise this, a fold will be slower too.

paf31:
Sorry this has gone unmerged for so long. Would you be interested in rebasing this on top of `master`? Or maybe this could be broken up into a few smaller PRs.

DavidHarrison:
Thanks! I'll split it up into smaller PRs.

paf31:
Ok thanks! I'll close this one then.

PR #39, opened 2015-08-16
Update license to include notice from GHC base

### LINKED ISSUE #37: LICENSE needs to credit Haskell base authors
Since `sort` is now taken from `base`. /cc @garyb

garyb:
Resolves #37 (oops).

@paf31 How does this look? Do we need to clarify anywhere else, do you think?


paf31:
:+1: Thanks.

Nowhere else that I know of right now.

PR #40, opened 2015-08-17
Add generic deriving

carymrobbins:


paf31:
:+1:

garyb:
@paf31 Would you say this is a breaking change?

I mean, it isn't, but the new version won't work in older version of `psc`.

paf31:
I think it should be a conscious decision to upgrade, and a major bump is the only way we have to ensure that, so yeah.

garyb:
In that case:

- [ ] purescript-quickcheck
- [ ] purescript-maps
- [ ] purescript-catenable-lists
- [ ] purescript-free
- [ ] purescript-sets
- [ ] purescript-graphs
- [ ] purescript-machines
- [ ] purescript-strongcheck
- [ ] purescript-argonaut-core
- [ ] purescript-argonaut-codecs
- [ ] purescript-argonaut-traversals
- [ ] purescript-argonaut

I'm not sure how comprehensive this tool is that I use to generate these, there must be some kind of bug in it as when I was doing the `strings` update it didn't list everything. I really should rewrite it in PS and get it doing all that other stuff we talked about...

garyb:
I'd suggest we add more derived generic instances and release them all as a batch if we're going down this route though, to prevent having to bump a bunch more times.

paf31:
Agreed, but what else needs a generic instance? I don't know if one for `Map` is appropriate for example. Anything with a `Generic` instance necessarily leaks its internal representation.

garyb:
Good point, perhaps there isn't anything else that would need it :) Let's just be sure before we commit!

paf31:
We have various newtypes which might benefit from it. In `purescipt-maybe` for example.

paf31:
Although of course that module can't depend on `generics` :smile:

garyb:
Good point, `monoids` has a bunch more too.

carymrobbins:
So, in regards to `Map`, we could just implement a Generic instance that doesn't expose its internal representation.  It could defer to `fromList`, see this commit - https://github.com/carymrobbins/purescript-maps/commit/5f8c41e94b26a7e463af06794f4653ae0107300c

paf31:
But then, e.g. `gShow` would be `show` on lists, which might be a bit odd.

carymrobbins:
That's how it basically is now, except for the qualified names and the extra, yet proper, parens -

```purescript
> singleton "foo" 1
fromList Cons (Tuple ("foo") (1)) (Nil)

-- With Generic (Map k v)
> data W = W (Map String Int)
> derive instance genericW :: Generic W
> instance showW :: Show W where show = gShow
> W (singleton "foo" 1)
W (Data.Map.Map.fromList (Cons (Data.Tuple.Tuple "foo" 1) Nil))

```

paf31:
@garyb Do we want to include this in 1.0? It seems harmless enough to me.

garyb:
Yeah, I think so!

PR #42, opened 2015-08-28
Make functions in Data.List stack-safe

zudov:
I've noticed, that some of the functions which certainly can be tail recursive aren't, so I've fixed the most obvious. But before going further I thought it would be wise to ask if there is some reason why they weren't tail recursive before?

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -185,8 +185,7 @@ null _ = false
 -- |
 -- | Running time: `O(n)`
 length :: forall a. List a -> Int
-length Nil = 0
-length (Cons _ xs) = 1 + length xs
+length = foldl (\acc _ -> acc + 1) 0
 
 --------------------------------------------------------------------------------
 -- Extending arrays ------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


garyb:
The `foldl` case is a compiler bug - purescript/purescript#1413. As for the others, yeah I think it's a mistake, they should be implemented as tail recursive.

paf31:
:+1: Looks great.

src/Data/List.purs
```diff
@@ -712,10 +711,10 @@ instance functorList :: Functor List where
...
   foldr _ b Nil = b
   foldr o b (Cons a as) = a `o` foldr o b as
-  foldl _ b Nil = b
-  foldl o b (Cons a as) = foldl o (b `o` a) as
-  foldMap _ Nil = mempty
-  foldMap f (Cons x xs) = f x <> foldMap f xs
+  foldl = go
+    where go _ b Nil = b
+          go o b (Cons a as) = go o (b `o` a) as
+  foldMap f = foldl (\acc -> append acc <<< f) mempty
```
paf31: This is left-to-right now, but was right-to-left before. For non-commutative monoids, that might be an issue.

paf31: Oh wait, maybe not, because `acc` is appended first. Could you please add a test for `foldMap` with `String`s?

zudov: @paf31 Done

zudov:
I am trying to finish it up and can't think of a nice way to implement `foldr`, is there one?

paf31:
No I don't think so, not without trampolining.

garyb:
What about implementing it in terms of `foldl`?

``` purescript
foldr f a bs = foldl (\g b x -> g (f b x)) id bs a
```

zudov:
@garyb Hmm. I don't remember if I tried that, but looks promising. I'll try soon.

paf31:
I think you'll just build up a tower of functions and blow the stack that way instead.

garyb:
Yeah, you could be right, I didn't try it aside from type checking.

paf31:
:+1: Thanks!

PR #44, opened 2015-08-31
Added Generic instance

zudov:


zudov:
Ping. Can we merge this?

zudov:
Uups. Just found out that it's a duplicate of #40

PR #46, opened 2015-09-12
Make show return "toList [1, 2, 3]" instead of "(Cons 1 (Cons 2 (Cons 3 Nil)))"

rightfold:
Fixes https://github.com/purescript/purescript-lists/issues/43

### DIFF
--- a/src/Data/List/Lazy.purs
+++ b/src/Data/List/Lazy.purs
@@ -629,10 +629,7 @@ zip = zipWith Tuple
 --------------------------------------------------------------------------------
 
 instance showList :: (Show a) => Show (List a) where
-  show xs = "fromStrict (" ++ go (step xs) ++ ")"
-    where
-    go Nil = "Nil"
-    go (Cons x xs) = "Cons (" ++ show x ++ ") (" ++ go (step xs) ++ ")"
+  show xs = "toList " <> show (fromList xs :: Array a)
 
 instance eqList :: (Eq a) => Eq (List a) where
   eq xs ys = go (step xs) (step ys)
--- a/src/Data/List/Lazy.purs
+++ b/src/Data/List/Lazy.purs
@@ -629,7 +629,10 @@ zip = zipWith Tuple
 --------------------------------------------------------------------------------
 
 instance showList :: (Show a) => Show (List a) where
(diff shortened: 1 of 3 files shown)


src/Data/List/Lazy.purs
```diff
@@ -629,10 +629,7 @@ zip = zipWith Tuple
 --------------------------------------------------------------------------------
 
 instance showList :: (Show a) => Show (List a) where
-  show xs = "fromStrict (" ++ go (step xs) ++ ")"
-    where
-    go Nil = "Nil"
-    go (Cons x xs) = "Cons (" ++ show x ++ ") (" ++ go (step xs) ++ ")"
+  show xs = "toList " <> show (fromList xs :: Array a)
```
paf31: Perhaps this is a good time to add a `take` with an ellipsis here to avoid trying to print infinite lazy lists?

paf31:
:+1:

src/Data/List/Lazy.purs
rightfold: How many elements to show by default?

rightfold: Let me add a unit test for that too before merging.

garyb:
:+1:

src/Data/List/Lazy.purs
```diff
@@ -629,10 +629,10 @@ zip = zipWith Tuple
 --------------------------------------------------------------------------------
 
 instance showList :: (Show a) => Show (List a) where
-  show xs = "fromStrict (" ++ go (step xs) ++ ")"
-    where
-    go Nil = "Nil"
-    go (Cons x xs) = "Cons (" ++ show x ++ ") (" ++ go (step xs) ++ ")"
+  show xs = "toList " <> show (fromList truncated :: Array a) <> ellipsis
+    where truncated = take 5 six
```
paf31: Sorry, I'm going to be a pain here. I just thought about this a bit more and realized, this way, there's no way to `show` more than 5 elements. Maybe we should create another function `showPrefix :: forall a. (Show a) => Int -> List a -> String`? That way, `show` is unsafe on infinite lists, but things stay flexible.

rightfold:
I'm redoing this since the array-specific functions are gone and there's conflicts.

PR #49, opened 2015-09-27
Import Data.Unfoldable explicitly to avoid name collision

spicydonuts:
Fixes https://github.com/purescript/purescript-lists/issues/48


paf31:
Thanks!

PR #51, opened 2015-12-14
Improve names for toList, fromList.

### LINKED ISSUE #50: Consider renaming fromList, toList
I almost always import `Data.List` qualified, like this:

```
import Data.List (List())
import Data.List as List
```

I think this is the best approach, as it makes the code easier to read (you don't have to remember which functions are in scope), and it means that you don't have to be tweaking your imports all the time (which is annoying and decreases the signal-to-noise ratio of diffs).

However it doesn't work very well with `toList` and `fromList`. I think these would be better named `fromFoldable` and `toUnfoldable` respectively; I think they convey the meaning better, and also `List.toList` is a bit redundant. See also purescript/purescript-maps#42.

hdgarrood:
Fixes #50

### DIFF
--- a/docs/Data/List.md
+++ b/docs/Data/List.md
@@ -25,39 +25,39 @@ which case it consists of a head element, and another list (represented by the
 
 ##### Instances
 ``` purescript
-instance showList :: (Show a) => Show (List a)
-instance eqList :: (Eq a) => Eq (List a)
-instance ordList :: (Ord a) => Ord (List a)
-instance semigroupList :: Semigroup (List a)
-instance monoidList :: Monoid (List a)
-instance functorList :: Functor List
-instance foldableList :: Foldable List
-instance unfoldableList :: Unfoldable List
-instance traversableList :: Traversable List
-instance applyList :: Apply List
-instance applicativeList :: Applicative List
-instance bindList :: Bind List
-instance monadList :: Monad List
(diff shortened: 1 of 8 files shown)


garyb:
:+1:!

paf31:
:+1: but I'm not looking forward to the downstream updates needed :smile: 

Does it make sense to turn these into aliases? Or is the issue that the names tend to collide with other things?

hdgarrood:
I haven't personally had problems with name collisions, I'm more interested in code readability. We could do aliases, although I would like to take advantage of being <1.0 while we still can - having both would feel like a wart. I can also volunteer for the downstream updates. :)

hdgarrood:
Speaking of which - does anyone have the incantation to list downstream libraries?

garyb:
Maybe hold off on that? I have a thing I'm working on with a post-0.8-release of the Prelude that should probably be a released as a breaking-version to prevent our previous disasters with version problems after compiler updates.

hdgarrood:
Ah, ok. So maybe we should leave them in as aliases for now then?

hdgarrood:
I've just amended this commit so that toList and fromList are still there as aliases, so that this can be merged without a major version bump. This way, people can start writing future-proof code now.

paf31:
Perfect, thanks!

PR #55, opened 2016-01-06
Make unfoldr stack-safe.

rgrempel:
The `unfoldableList` implementation wasn't tail-recursive, which caused the problems described here:

https://github.com/hdgarrood/purescript-sequences/issues/20

This pull request changes `unfoldr` so that it is tail-recursive. The provided test fails without the change.

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -733,10 +733,11 @@ instance foldableList :: Foldable List where
 
 
 instance unfoldableList :: Unfoldable List where
-  unfoldr f b = go (f b)
+  unfoldr f b = go b Nil
     where
-    go Nothing = Nil
-    go (Just (Tuple a b)) = Cons a (go (f b))
+      go source memo = case f source of
+        Nothing -> reverse memo
+        Just (Tuple one rest) -> go rest (Cons one memo)
 
 instance traversableList :: Traversable List where
   traverse _ Nil = pure Nil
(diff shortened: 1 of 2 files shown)


src/Data/List.purs
```diff
@@ -733,10 +733,11 @@ instance foldableList :: Foldable List where
 
 
 instance unfoldableList :: Unfoldable List where
-  unfoldr f b = go (f b)
+  unfoldr f b = go b Nil
     where
-    go Nothing = Nil
-    go (Just (Tuple a b)) = Cons a (go (f b))
+      go source memo = case f source of
+        Nothing -> memo
```
paf31: Doesn't this need to be reversed?

rgrempel:
I have now noticed that my implementation reverses the list ... I'll work on fixing that :-)

rgrempel:
I've added the call to `reverse`, and I squashed the commits.

I wasn't sure whether the right thing to do was reverse the whole thing at the end (which is what I chose), or try to add each individual item to the end as we go. If the latter is better, I'd be happy to do that instead.

rgrempel:
I took a look at `Data.List.snoc`, and it seems clear that it is better to build up the list backwards and then reverse it at the end (as this PR now does) -- otherwise, one would be reversing and re-reversing a lot!

paf31:
:+1:  Thanks!

PR #56, opened 2016-01-07
adds mapWithIndex function.

kRITZCREEK:
This is the simplest implementation I could think of. Please check if there are performance troubles.

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -51,6 +51,7 @@ module Data.List
   , filterM
   , mapMaybe
   , catMaybes
+  , mapWithIndex
 
   , sort
   , sortBy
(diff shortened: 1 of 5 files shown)


src/Data/List.purs
```diff
@@ -428,6 +429,14 @@ mapMaybe f = go Nil
 catMaybes :: forall a. List (Maybe a) -> List a
 catMaybes = mapMaybe id
 
+
+-- | Apply a function to each element and its index in a list starting at 0.
+mapWithIndex :: forall a b. (a -> Int -> b) -> List a -> List b
+mapWithIndex f = go 0
+  where
+  go _ Nil = Nil
+  go n (Cons a as) = Cons (f a n) (go (n+1) as)
```
paf31: If we could make this tail recursive, that would be nice.

paf31:
:+1: LGTM, thanks!

garyb:
:+1:

garyb:
Thanks!

PR #57, opened 2016-02-13
Fix warnings generated by purescript-0.8.0

telser:


src/Data/List.purs
```diff
@@ -194,14 +191,13 @@ length = foldl (\acc _ -> acc + 1) 0
...
-
 -- | An infix alias for `Cons`; attaches an element to the front of
 -- | a list.
 -- |
 -- | Running time: `O(1)`
-(:) :: forall a. a -> List a -> List a
-(:) = Cons
+cons :: forall a. a -> List a -> List a
+cons = Cons
+infixr 6 cons as :
```
Thimoteus: With this definition for (:) I get the following error when trying to use it to pattern match (using 0.8.2):

`Operator (:) cannot be used in a pattern as it is an alias for function cons.
  Only aliases for data constructors may be used in patterns.`

garyb: Yeah, as of v0.8.2 this should be `infixr 6 Cons as :` instead :+1:

telser: Ah yes, will fix. Thanks for pointing it out.

nullobject:
Nice :+1:

garyb:
Thanks!

PR #58, opened 2016-02-21
Add Data.List.ZipList.transpose

hdgarrood:
This seemed to be the most appropriate place I could think of for it. I also added some tests for ZipList.

### DIFF
--- a/src/Data/List/ZipList.purs
+++ b/src/Data/List/ZipList.purs
@@ -4,6 +4,7 @@
 module Data.List.ZipList
   ( ZipList(..)
   , runZipList
+  , transpose
   ) where
 
 import Prelude
@@ -13,9 +14,11 @@ import Control.Alternative (Alternative)
 import Control.Plus (Plus)
 
 import Data.Foldable (Foldable, foldMap, foldl, foldr)
-import Data.List.Lazy (List(), repeat, zipWith)
+import Data.List.Lazy (List(), repeat, zipWith, fromFoldable, toUnfoldable)
 import Data.Monoid (Monoid, mempty)
 import Data.Traversable (Traversable, traverse, sequence)
+import Data.Foldable (Foldable)
+import Data.Unfoldable (Unfoldable)
 
 -- | `ZipList` is a newtype around `List` which provides a zippy
 -- | `Applicative` instance.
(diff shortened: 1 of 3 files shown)


src/Data/List/ZipList.purs
```diff
@@ -25,6 +28,24 @@ newtype ZipList a = ZipList (List a)
 runZipList :: forall a. ZipList a -> List a
 runZipList (ZipList xs) = xs
 
+-- | Transpose any structure which is both `Foldable` and `Unfoldable`,
+-- | via `ZipList`; note that `sequence` for `ZipList` is the same as
+-- | `transpose`.
+transpose :: forall f a. (Foldable f, Unfoldable f) => f (f a) -> f (f a)
```
paf31: I think these overloaded functions have their place, but I'd prefer to expose a `List` version too, since we're going via that function anyway.

src/Data/List/ZipList.purs
```diff
@@ -25,6 +28,24 @@ newtype ZipList a = ZipList (List a)
...
+transpose =
```
paf31: Nice implementation :smile:

src/Data/List/ZipList.purs
hdgarrood: I thought about this, but what about name collisions? There's always straight `sequence` if you want to avoid all the other stuff, eg for performance.

paf31: I'm just concerned that a beginner won't know to use `sequence`, and may not understand the more general type signature.

hdgarrood: We already have generalized type signatures for all kinds of things, though? `foldr`, `replicate`, `(<>)`, even `(+)` and `(*)`.

Just so that we're on the same page - are you suggesting we put specialized versions of `transpose` into `Data.List` and `Data.List.Lazy`, as well as this generalized version here?

src/Data/List/ZipList.purs
jdegoes: In my opinion, such a highly generalized implementation as the above `transpose` does not belong here, because it has no dependency on `Data.List`. (One can argue it belongs in `Foldable` or `Unfoldable`.)

paf31: My point is just that there is a difference between

- a function which is initially written in terms of `List` and then generalized because `Foldable` is basically just `toList`, and
- a function which is by definition, already generalized to work with some type class.

`Foldable` is a particularly confusing case since it basically defines a monoid homomorphism to `List`. But in the first case, a user can always choose to easily generalize to `Foldable` using a call to `toList`.

hdgarrood: Given that I don't have to worry about the core library dependency graph, I won't make any suggestions about where I think this should go, I'll leave that up to you ;)

I will say, though, that I don't see how this could go into `foldable-traversable` or `unfoldable` (at least, not without duplicating parts of `lists`) because the implementation does depend on the `Traversable ZipList` instance, and we can't depend on `ZipList` in `foldable-traversable` or `unfoldable` because `lists` already depends on both of those.

Maybe we should just start with a version defined in terms of `List` and not `Foldable`, then, and see what happens.

Come to think of it, why does `ZipList` wrap lazy lists and not strict ones? Would it make sense to have a zippy strict list newtype too?

paf31: Because `pure` is infinite :smile:

hdgarrood: Ah, lol, of course.

paf31: Speaking of which, doesn't this diverge for empty input?

hdgarrood: Oops, yes it does.

hdgarrood:
See #59 for a second attempt.

PR #66, opened 2016-04-28
Pretty show

### LINKED ISSUE #43: Make show return "toList [1, 2, 3]" instead of "(Cons 1 (Cons 2 (Cons 3 Nil)))"
I think it would be beneficial to make `show` return `"toList [1, 2, 3]"` instead of `"(Cons 1 (Cons 2 (Cons 3 Nil)))"`, as that's much more readable.

rightfold:
Fixes #43. Supersedes #46.

### DIFF
--- a/src/Data/List/Lazy.purs
+++ b/src/Data/List/Lazy.purs
@@ -647,10 +647,7 @@ transpose xs =
 --------------------------------------------------------------------------------
 
 instance showList :: Show a => Show (List a) where
-  show xs = "fromStrict (" <> go (step xs) <> ")"
-    where
-    go Nil = "Nil"
-    go (Cons x xs) = "(Cons " <> show x <> " " <> go (step xs) <> ")"
+  show xs = "fromFoldable " <> show (toUnfoldable xs :: Array a)
 
 instance eqList :: Eq a => Eq (List a) where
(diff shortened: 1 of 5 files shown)


src/Data/List/Lazy.purs
```diff
@@ -643,14 +645,26 @@ transpose xs =
           (x : mapMaybe head xss) : transpose (xs : mapMaybe tail xss)
 
 --------------------------------------------------------------------------------
+-- Infinite list inspection ----------------------------------------------------
+--------------------------------------------------------------------------------
+
+-- | Like `show`, but only shows the first n elements. Useful for potentially
+-- | infinite lists.
+showPrefix :: forall a. (Show a) => Int -> List a -> String
```
paf31: I think if users need this, they can just write `show <<< take n`. I do this frequently in Haskell, for example.

paf31:
👍

src/Data/List/Lazy.purs
rightfold: Fair enough, I'll rebase it out

garyb:
I just merged #74 which had an alternative pretty-`show` using `:`, so closing this. This would have been equally good in my book, so sorry it never went in!

PR #68, opened 2016-05-18
Avoid using operators for binding arguments

hdgarrood:
It seems the compiler as of the current master (4852b5d) is no longer able to bind operators as function arguments. If this is intentional, should it be listed in https://github.com/purescript/purescript/issues/2012?


garyb:
Perhaps yeah, it's implied by "Remove non-aliased operator declarations" but that's probably not obvious.

Thanks!

PR #69, opened 2016-05-22
Add Generic instance

garyb:
Resolves #40 

Will probably fail to build because the new generics requires psc 0.9.

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -94,6 +94,7 @@ import Control.MonadZero (class MonadZero)
 import Control.Plus (class Plus)
 
 import Data.Foldable (class Foldable, foldl, foldr, any)
+import Data.Generic (class Generic)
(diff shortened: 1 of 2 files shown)


src/Data/List.purs
```diff
@@ -689,6 +690,8 @@ foldM f a (Cons b bs) = f a b >>= \a' -> foldM f a' bs
 -- Instances -------------------------------------------------------------------
 --------------------------------------------------------------------------------
 
+derive instance genericList :: Generic a => Generic (List a)
+
 instance showList :: Show a => Show (List a) where
   show Nil = "Nil"
```
paf31: Wasn't this changed to render something like `toList [1,2,3]`?

paf31: Oh never mind, it's in an open PR.

paf31:
👍 Looks good!

PR #73, opened 2016-06-08
Add `Fail` instance for `Bind ZipList`

### LINKED ISSUE #72: ZipList Bind/Monad
If I understand the new programmable type errors, it seems like this instance can be added:

```purescript
instance zipListIsNotBind
  :: Fail "ZipList is not Bind. Any implementation would break the associativity law"
  => Bind ZipList where
    bind = crashWith "bind: unreachable"
```

Also, does this stop the hierarchy? Or can someone take this `Bind` instance and use it to make a `Monad ZipList` and higher instances?

joneshf:
Closes #72.

### DIFF
--- a/src/Data/List/ZipList.purs
+++ b/src/Data/List/ZipList.purs
@@ -65,3 +65,8 @@ instance plusZipList :: Plus ZipList where
   empty = mempty
 
 instance alternativeZipList :: Alternative ZipList
+
+instance zipListIsNotBind
+  :: Fail "ZipList is not Bind. Any implementation would break the associativity law"
+  => Bind ZipList where
+    bind _ _ = mempty
--- a/src/Data/List/ZipList.purs
+++ b/src/Data/List/ZipList.purs
@@ -7,15 +7,14 @@ module Data.List.ZipList
   ) where
 
 import Prelude
-
 import Control.Alt (class Alt)
 import Control.Alternative (class Alternative)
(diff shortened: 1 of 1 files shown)


src/Data/List/ZipList.purs
```diff
@@ -65,3 +65,8 @@ instance plusZipList :: Plus ZipList where
   empty = mempty
 
 instance alternativeZipList :: Alternative ZipList
+
+instance zipListIsNotBind
+  :: Fail "ZipList is not Bind. Any implementation would break the associativity law"
```
paf31: It might make sense to suggest regular `List` here?

src/Data/List/ZipList.purs
```diff
@@ -65,3 +65,8 @@ instance plusZipList :: Plus ZipList where
...
+  => Bind ZipList where
+    bind _ _ = mempty
```
paf31: Can we use `unsafeCrashWith` to avoid confusion (in case anyone thinks this is runnable code)?

paf31:
Thanks!

joneshf:
Updated. Let me know if you'd like the error to be something else. Or if there's precedent somewhere, link me up :smile:

garyb:
New version looks nice! And I think you're setting the precedent here, we've not made use of this anywhere else yet. :smile:

paf31:
👍 LGTM

PR #74, opened 2016-07-08
Use colon operator instead of Cons

### LINKED ISSUE #43: Make show return "toList [1, 2, 3]" instead of "(Cons 1 (Cons 2 (Cons 3 Nil)))"

sharkdp:
- Use the infix operator instead of the `Cons` constructor to clean up the code, especially because it remove a lot of parenthesis.
- Modify the `Show` instance to use the infix operator instead of `Cons`:

``` purs
> show (1 : 2 : 3 : Nil)
"(1 : (2 : (3 : Nil)))"
```

Personally, I like this better than

``` purs
"(Cons 1 (Cons 2 (Cons 3 Nil)))"
```

Maybe it would even close #43.

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -93,7 +93,7 @@ import Control.MonadPlus (class MonadPlus)
 import Control.MonadZero (class MonadZero)
 import Control.Plus (class Plus)
 
-import Data.Foldable (class Foldable, foldl, foldr, any)
+import Data.Foldable (class Foldable, foldl, foldr, any, intercalate)
 import Data.Generic (class Generic)
 import Data.Maybe (Maybe(..))
 import Data.Monoid (class Monoid, mempty)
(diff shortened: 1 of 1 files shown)


paf31:
👍 Looks good.

I think it would be nice to use `intercalate ":"` in the `show` instance to reduce the unnecessary parens though, if possible.

sharkdp:
> I think it would be nice to use intercalate ":" in the show instance to reduce the unnecessary parens though, if possible.

Oh right, this is much better. Now it looks like this:
``` purs
> show (1 : 2 : 3 : Nil)
"(1 : 2 : 3 : Nil)"
```

paf31:
👍  Thanks!

garyb:
Thanks!

PR #77, opened 2016-08-07
add `replicate` and `replicateM` to Data.List

rvion:
Those functions were missing.

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -14,6 +14,8 @@ module Data.List
 
   , singleton
   , (..), range
+  , replicate
+  , replicateM
   , some
   , many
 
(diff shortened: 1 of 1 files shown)


sharkdp:
There as generalized versions of these in purescript-unfoldable:
[https://github.com/purescript/purescript-unfoldable/blob/master/src/Data/Unfoldable.purs](https://github.com/purescript/purescript-unfoldable/blob/master/src/Data/Unfoldable.purs)

garyb:
Yiup! They were moved out to `Unfoldable` in the latest round of library updates.

rvion:
arf, should have thought about it, Thanks !

any reason for `replicate` and `replicateM` to be in Data.List.Lazy 
(to take advantage of repeat ?)

PR #78, opened 2016-10-08
Updates for 2.0

### LINKED ISSUE #75: Use NonEmpty for group, group' and groupBy?
Using purescript-nonempty, the signature of `group` could be changed from
``` purs
group :: forall a. Eq a => List a -> List (List a)
```
to
``` purs
group :: forall a. Eq a => List a -> List (NonEmpty List a)
```

I believe this is not just more descriptive, but also useful because it helps people to avoid `unsafePartial <<< Data.List.Partial.head` or similar constructions.

On the downside, it pulls in the purescript-nonempty dependency.

The same applies to `group'` and `groupBy`.

paf31:
Blocked by `generics`, and we should probably merge some of the open PRs first too. 

Resolves #75

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,7 +24,8 @@
     "purescript-foldable-traversable": "^2.0.0",
     "purescript-generics": "v2.0.0",
     "purescript-lazy": "^2.0.0",
-    "purescript-unfoldable": "^2.0.0"
+    "purescript-unfoldable": "^2.0.0",
+    "purescript-nonempty": "^2.0.0"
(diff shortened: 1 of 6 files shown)


garyb:
Should we go back and make the `NonEmpty` changes to arrays too?

paf31:
Yes, I think that's a good idea.

garyb:
I think it's done at last :smile:

paf31:
👍 Looks great!

PR #79, opened 2016-10-13
Add NonEmptyList basics

garyb:


src/Data/List/NonEmpty.purs
```diff
@@ -0,0 +1,84 @@
...
+head :: forall a. NonEmptyList a -> a
+head (NonEmptyList (x :| _)) = x
+
+last :: forall a. NonEmptyList a -> a
+last (NonEmptyList (x :| xs)) = fromMaybe x (L.last xs)
+
+tail :: forall a. NonEmptyList a -> L.List a
+tail (NonEmptyList (_ :| xs)) = xs
+
+init :: forall a. NonEmptyList a -> NonEmptyList a
```
paf31: The result isn't non-empty in general though. This gives the equivalent of `init [x] = [x]`.

src/Data/List/NonEmpty.purs
```diff
@@ -0,0 +1,84 @@
...
+  apply (NonEmptyList (f :| fs)) (NonEmptyList (a :| as)) =
+    NonEmptyList (f a :| (fs <*> L.singleton a) <> ((f : fs) <*> as))
+
+instance applicativeNonEmptyList :: Applicative NonEmptyList where
+  pure = NonEmptyList <<< NE.singleton
+
+instance bindList :: Bind NonEmptyList where
+  bind = flip concatMap
+
+instance monadNonEmptyList :: Monad NonEmptyList
```
paf31: We can add a `Comonad` instance too (later maybe?)

src/Data/List/NonEmpty.purs
garyb: Ah yes, thanks!

src/Data/List/NonEmpty.purs
garyb: For sure

src/Data/List/NonEmpty.purs
```diff
@@ -0,0 +1,84 @@
...
+uncons (NonEmptyList (x :| xs)) = { head: x, tail: xs }
+
+length :: forall a. NonEmptyList a -> Int
+length (NonEmptyList (x :| xs)) = 1 + L.length xs
+
+concatMap :: forall a b. (a -> NonEmptyList b) -> NonEmptyList a -> NonEmptyList b
+concatMap f (NonEmptyList (a :| as)) =
+  case f a of
+    NonEmptyList (b :| bs) ->
+      NonEmptyList (b :| bs <> L.concatMap (NE.oneOf <<< unwrap <<< f) as)
```
paf31: Instead of using `oneOf <<< unwrap`, it might be good to add `toList` which is probably going to be quite a bit faster.

paf31:
👍 Thanks for adding this

src/Data/List/NonEmpty.purs
garyb: Good idea. I guess I was trying to resist that since we dropped all the specialised `toLists` in favour of `fromFoldable`, but it makes sense here I think!

garyb:
Can you check my `extend` implementation? I think it's right... I wanted to avoid a recursive version, since it's not tail recursive.

garyb:
We can add the `Extend` instance for the normal `List` too I realised while doing this, just not `Comonad`.

paf31:
It looks right at first glance, but (just a thought) - it might be easier to write `tails` (which is `duplicate`) as a fold over the non-empty list itself.

garyb:
Yeah, I considered that, but I figured it wouldn't clean things up much, as it'd be the same implementation with just the `f` taken out I think.

paf31:
Pretty much 😄

paf31:
Also, I think this can be written as a `scanr` although it might be slower.

garyb:
Ok, I'm done with this now. If you're happy with it as is I'll get it out and update `foreign`.

garyb:
Ah yeah, I think `scanr` would help with a `duplicate` - but then I wasn't sure about the "leaves the initial item out" thing.

PR #81, opened 2016-10-15
Use NonEmptyList rather than NonEmpty List for grouping

garyb:
Sooo... it's kinda weird and annoying that the `group` functions return `NonEmpty List` rather than `NonEmptyList`, since unless you then `wrap` stuff, you can't take advantage of the NEL operations.

This does introduce a circular dependency though, so, I have a `Types` module that has just the types & instances, which are then re-exported from the other places.

There are some other changes in here too:
- `foldr` for `List` is now implemented in terms of `foldl` so it is stack safe
- `map` is implemented with `foldr` rather than `reversing` a result from a left folding recursive function, as I wanted to avoid bringing `reverse` into the `Types` module

What do you think?

I guess we should probably add a `Lazy` version of `NonEmptyList` so we can do the same thing there too...


jdegoes:
Another formulation worth considering:

```haskell
type NonEmptyList a = Cofree Maybe a
```

I believe this is already lazy as `Cofree` is lazy. It has a slight advantage in that the "tail" of the list has the same structure (if it's not Nothing), rather than a different structure as for `NonEmpty List` (in which you can use NonEmptyList functions on the top level, but after you dig into that, you have to use List functions).

src/Data/List/Types.purs
```diff
@@ -0,0 +1,167 @@
...
+  append (x : xs) ys = x : (xs <> ys)
+
+instance monoidList :: Monoid (List a) where
+  mempty = Nil
+
+instance functorList :: Functor List where
+  map f = foldr (\x acc -> f x : acc) Nil
+
+instance foldableList :: Foldable List where
+  foldr f b as = foldl (\g a x -> g (f a x)) id as b
```
natefaubion: If you are folding with continuations, how is this stack safe? The folding may be stack safe, but running the composition of continuations will blow the stack won't it (there is no test for the stack safety of foldr)?

paf31: True, what about using `reverse` and `foldl` to define `foldr`?

paf31:
👍 LGTM, excluding the `foldr` issue. Is the plan to release another set of major revisions for any reverse dependencies?

paf31:
I'd also be interested to consider mutually recursive definitions of `NonEmptyList` and `List`, which is similar to @jdegoes' suggestion without pulling in `free` as a dependency. It would be interesting to try to write (non-mutually) recursive functions on lists with that approach though.

garyb:
Yeah, I figured doing another round of updates after this. The dependents list isn't too bad:

- purescript-quickcheck
- purescript-foreign
- purescript-catenable-lists
- purescript-semirings
- purescript-maps

I think there's a few transitive dependencies also, like `free`.

garyb:
The alternative `NonEmptyList` representation definitely has some appeal, but I think there's also some benefit in the current approach since it preserves the trivial relationship between `NonEmpty List` and `NonEmptyList`. Although I guess that could still be done with an iso, just there'd be an O(1) conversion overhead there.

The `Lazy` version of `NonEmptyList` is strict in the head with this `NonEmpty List` representation - we could perhaps avoid that with an alternative formulation, but I'm not sure how big a deal it is?

garyb:
Oh, I added `Extend` and `Newtype` instances for the lazy list in the last commit too.

paf31:
👍 Looks good, but I'm just a little wary of letting the lazy non-empty list be strict in the head element. I guess we can see if it becomes an issue, but perhaps hiding the constructor for now is a good way forward?

garyb:
How about this version then? (`newtype NonEmptyList a = NonEmptyList (Lazy (NonEmpty List a))`)

paf31:
👍 Looks great, thanks!

PR #82, opened 2016-10-16
Use NonEmptyList for lazy version of group functions

garyb:
Nearly made a stupid mistake there... deleted the v3 release immediately after tagging it, so will re-release as that once this is in.


paf31:
👍

PR #84, opened 2016-10-28
Fix `-generics` dependency

garyb:
I don't know what happened here!


paf31:
👍 LGTM. Does this need a new release?

garyb:
Yeah, I'll just make it a patch though - it's currently blocking anything that requires generics v3.1.0.

PR #85, opened 2016-10-28
Add `MonadRec` versions of `many` and `some`

garyb:
I don't think these versions need the `Lazy` constraint?


paf31:
👍 Looks good, thanks!

PR #89, opened 2016-11-21
Make `append` stack-safe

sammthomson:
Re-implement `append` in terms of `foldr` to avoid
"RangeError: Maximum call stack size exceeded" errors.
Add tests for functionality and stack-safety.


sammthomson:
fixes Issue #90

paf31:
Thanks!

Issue #91, opened 2016-11-26
Data.List.Lazy has several stack-unsafe methods


sammthomson:
`Data.List.Lazy` has several stack-unsafe methods: `length`, `reverse`, `foldr`, `foldl`, `foldMap`, and `traverse`. They have simple fixes and I have an incoming PR.

`foldr` and `traverse` are things where it seems like a lazy version would be useful. But as far as I can tell it's impossible to implement them lazily without changing their types in [purescript-foldable-traversable](https://github.com/purescript/purescript-foldable-traversable). E.g., a lazy foldr might have a type like
```purescript
foldrLazy :: forall a b. Lazy b => (a -> b -> b) -> b -> f a -> b
```
Since they can't be lazy given their current types, they might as well be stack-safe.

PR #92, opened 2016-11-26
Fixing several stack-unsafe functions in Data.List.Lazy

sammthomson:
Fixes issue #91.
Also made `unfoldr` and `toList` slightly lazier.


paf31:
Thanks again!

PR #93, opened 2016-12-26
Use ~>

mlang:


jdegoes:
👍 Nice. I really like this style and I'd be happy to see it more pervasive in the std libraries.

PR #94, opened 2017-01-20
Add unsnoc

joshuahhh:


paf31:
👍 Looks good to me, thanks!

src/Data/List.purs
```diff
@@ -257,6 +258,13 @@ uncons :: forall a. List a -> Maybe { head :: a, tail :: List a }
 uncons Nil = Nothing
 uncons (x : xs) = Just { head: x, tail: xs }
 
+-- | Break a list into its last element, and the preceding elements,
+-- | or `Nothing` if the list is empty.
+-- |
+-- | Running time: `O(2n)`
+unsnoc :: forall a. List a -> Maybe { init :: List a, last :: a }
+unsnoc xs = { init: _, last: _ } <$> init xs <*> last xs
```
paf31: Actually, can we do this in a single pass? Otherwise, the user doesn't gain anything over using the functions separately.

joshuahhh: My motivation for adding it was that it makes pattern-matching cleaner. Otherwise, you have to do separate `Maybe` checks for both `init` and `last`. But sure, I'd be happy to optimize it. 😃

joshuahhh:
`unsnoc` now unsnocs in a single pass, rather than using `init` and `last` separately. It is based on the implementation of `init`. Since it is exactly as performant as `init`, `init` is now defined in terms of `unsnoc`. `last`, on the other hand, avoids the call to `reverse`, so I left it the way it is.

src/Data/List.purs
```diff
@@ -257,6 +253,18 @@ uncons :: forall a. List a -> Maybe { head :: a, tail :: List a }
...
 uncons (x : xs) = Just { head: x, tail: xs }
 
+-- | Break a list into its last element, and the preceding elements,
+-- | or `Nothing` if the list is empty.
+-- |
+-- | Running time: `O(n)`
+unsnoc :: forall a. List a -> Maybe { init :: List a, last :: a }
+unsnoc lst = unsnocHelper lst Nil <#> \h -> { init: reverse h.revInit, last: h.last }
+
+unsnocHelper :: forall a. List a -> List a -> Maybe { revInit :: List a, last :: a }
```
paf31: Could you please move this under a `where`? It could be called `go` - it's not the best name, but it's consistent with the other implementations here.

joshuahhh: Sure! The main reason I split it off was concerns about performance – I figured that redefining `go` on every call to `unsnoc` was wasteful. But I don't know much about that, so I'll take your advice.

paf31: I don't know the extent to which JS engines optimize under functions, but I'd say it's better to be consistent at this point.

paf31:
👍 Thanks!

PR #97, opened 2017-02-02
sorting for NonEmptyList

matthewleon:
Adds the two sorting functions to `NonEmptyList`.

### DIFF
--- a/src/Data/List/NonEmpty.purs
+++ b/src/Data/List/NonEmpty.purs
@@ -13,6 +13,8 @@ module Data.List.NonEmpty
   , length
   , concatMap
   , appendFoldable
+  , sort
+  , sortBy
   ) where
 
 import Prelude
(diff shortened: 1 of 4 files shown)


paf31:
@garyb  I think we can make this a minor version bump, yes?

src/Data/List/NonEmpty.purs
```diff
@@ -62,3 +65,10 @@ length (NonEmptyList (x :| xs)) = 1 + L.length xs
 
 concatMap :: forall a b. (a -> NonEmptyList b) -> NonEmptyList a -> NonEmptyList b
 concatMap = flip bind
+
+sort :: forall a. Ord a => NonEmptyList a -> NonEmptyList a
+sort xs = sortBy compare xs
+
+sortBy :: forall a. (a -> a -> Ordering) -> NonEmptyList a -> NonEmptyList a
+sortBy cmp xs = unsafeFromList $ L.sortBy cmp (toList xs)
```
paf31: Another option here is to sort the tail and insert the head into the sorted result, a la insertion sort. That would require us to change the type of `insert` though, or to provide an alternative version.

paf31:
@matthewleon Sorry for the delay, could you please rebase?

garyb:
> I think we can make this a minor version bump, yes?

Agreed 👍

src/Data/List/NonEmpty.purs
matthewleon: @paf31 are you suggesting we should go down that route? I understand that `unsafeFromList` is, in many cases, a last resort, but here it seems to be both safe and more efficient than the alternative.

garyb: I'd vote for the `unsafeFromList` version here too, personally. :)

matthewleon:
> Sorry for the delay, could you please rebase?

done

paf31:
Thanks!

PR #102, opened 2017-02-06
[BREAKING] Swap arguments to function in `mapWithIndex`

garyb:
For consistency with `purescript-arrays`. Or we could do it the other way around? I don't have a strong preference for the ordering, but figured if I was doing it in Haskell I'd produce a list of tuples with the index first.


paf31:
👍

PR #103, opened 2017-02-13
appendList function for nonempty lists

matthewleon:
For both strict and lazy nonempty lists, a function
appendList :: NonEmptyList ~> List ~> NonEmptyList

(a very handy function for creating an Arbitrary instance for NonEmptyLists).


matthewleon:
actual type sig:
`appendList :: forall a. NonEmptyList a -> List a -> NonEmptyList a`

matthewleon:
On further thought, this is not really necessary for the Arbitrary instance. A handy function all the same, though.

garyb:
Perhaps we could generalize it to append anything that is `Foldable`?

garyb:
Thanks!

PR #113, opened 2017-04-26
add some and many for Lazy List

### LINKED ISSUE #112: add some many to Lazy List
There is no definition of some and many for Lazy List.
Is it intentional? (why)
If it's not i could open PR

safareli:
fix #112

### DIFF
--- a/src/Data/List/Lazy.purs
+++ b/src/Data/List/Lazy.purs
@@ -16,8 +16,8 @@ module Data.List.Lazy
   , (..), range
   , replicate
   , replicateM
-  -- , some
-  -- , many
+  , some
+  , many
   , repeat
   , iterate
   , cycle
@@ -91,6 +91,8 @@ module Data.List.Lazy
 
 import Prelude
 
+import Control.Alt ((<|>))
+import Control.Alternative (class Alternative)
 import Control.Lazy as Z
 
 import Data.Foldable (class Foldable, foldr, any, foldl)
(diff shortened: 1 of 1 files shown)


garyb:
Looks good to me :+1:

src/Data/List/Lazy.purs
```diff
@@ -160,6 +162,21 @@ replicateM n m
...
+-- |
+-- | The `Lazy` constraint is used to generate the result lazily, to ensure
+-- | termination.
+some :: forall f a. Alternative f => Z.Lazy (f (List  a)) => f a -> f (List  a)
+some v = cons <$> v <*> Z.defer (\_ -> many v)
+
+-- | Attempt a computation multiple times, returning as many successful results
+-- | as possible (possibly zero).
+-- |
+-- | The `Lazy` constraint is used to generate the result lazily, to ensure
```
paf31: There is no `Lazy` constraint here.

paf31:
Thanks!

src/Data/List/Lazy.purs
safareli: Copied it from [List.purs](https://github.com/purescript/purescript-lists/blob/master/src/Data/List.purs#L158).  But even tho `Lazy` constraint isn't used in `many`, it's used in `some`, which is used by `many`, so it's technically true.  If want better wording here please let me know and i'll update it in both places

paf31: Sorry, I misread it, please ignore my comment.

src/Data/List/Lazy.purs
```diff
@@ -160,6 +162,21 @@ replicateM n m
       as <- replicateM (n - one) m
       pure (cons a as)
 
+-- | Attempt a computation multiple times, requiring at least one success.
+-- |
+-- | The `Lazy` constraint is used to generate the result lazily, to ensure
+-- | termination.
+some :: forall f a. Alternative f => Z.Lazy (f (List  a)) => f a -> f (List  a)
+some v = cons <$> v <*> Z.defer (\_ -> many v)
```
paf31: Is it possible to generate the tail lazily here?

safareli: That's what `Z.defer` is doing here.

paf31: Yes, I think this is fine. Sorry again :)

safareli: It's fine :p

PR #115, opened 2017-05-12
Reimplement snoc with foldr

LiamGoodacre:
Reversing twice means we rebuild the entire list twice.  Also `O(2n) = O(n)`.


paf31:
👍 LGTM, thanks!

PR #117, opened 2017-05-30
partition

matthewleon:
Returns a `Tuple` as in Haskell. See https://github.com/purescript/purescript-arrays/issues/102

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -65,6 +65,7 @@ module Data.List
   , group
   , group'
   , groupBy
+  , partition
 
   , nub
   , nubBy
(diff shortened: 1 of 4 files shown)


matthewleon:
Just realized that there was a commented-out previous attempt at this. Have removed it in this PR.

hdgarrood:
Most core libraries do tend to prefer records over tuples, as it is much easier to remember which part of the result is which this way. Thanks for this but I think I prefer it how it is now.

hdgarrood:
Oh, didn't realise this function doesn't exist here yet! Okay yeah looks good except I think it would be better to use the same style as in arrays currently.

matthewleon:
Okay, I'll adjust this accordingly. Thanks.

matthewleon:
Should be good now.

michaelficarra:
`true` and `false` might be better than `yes` and `no`.

matthewleon:
> true and false might be better than yes and no.

I am inclined to agree, but `purescript-arrays` sets a precedent here that I'm following. Are there any other packages that use a true/false convention?

LiamGoodacre:
Btw what's the reason behind using reverse and foldl for strict lists?

Also, these were also added for lists in [purescript-filterable](https://github.com/LiamGoodacre/purescript-filterable/blob/f526723f481241f5a0308dcb04bebb20adb5063c/src/Data/Filterable.purs#L133) not long ago :)

garyb:
> Btw what's the reason behind using reverse and foldl for strict lists?

Stack safety, if I remember rightly.

LiamGoodacre:
Oh, looks like foldr is also defined by reverse and foldl 😄 https://github.com/purescript/purescript-lists/blob/master/src/Data/List/Types.purs#L68

matthewleon:
> Btw what's the reason behind using reverse and foldl for strict lists?
> 
> Also, these were also added for lists in purescript-filterable not long ago :)

Thanks @LiamGoodacre for this. I've basically copied your work over from `purescript-filterable`. I forgot about `foldr` being implemented in terms of `foldl` anyway, so it seems to me that your take on this one is indeed more elegant.

paf31:
Thanks!

Issue #111, opened 2017-04-07
use naive Lazy List foldr


matthewleon:
Currently, Data.List.Lazy defines `foldr` in terms of `foldl` on a `reverse`d list. This preserves stack safety.

Might it be worth sacrificing stack safety to allow `foldr` to work with potentially very large or infinite lists by using the naive definition (trusting the user to use a lazy operation)? If not, is that worth having as a separate function?

Quick example of a consequence of the current def: a function like `traverse_` has to wait until the entire list is traversed before actions can begin.

paf31:
Having a variant of `foldr` which looks like `(a -> Lazy b -> b) -> List a -> b -> b` can be very useful for modularity, or if you're using infinite or lazy lists. Let's add it!

matthewleon:
In the above did you intend to type `forall a b. (a -> Lazy b -> b) -> b -> List a -> b` or are arguments flipped on purpose?

matthewleon:
> Having a variant of foldr which looks like (a -> Lazy b -> b) -> List a -> b -> b can be very useful for modularity, or if you're using infinite or lazy lists. Let's add it!

In the above are arguments flipped on purpose or did you intend `(a -> Lazy b -> b) -> b -> List a -> b`.

paf31:
@matthewleon Shall

PR #118, opened 2017-05-31
foldr': lazy right folds

matthewleon:
addresses #111

This looks a bit different from the signature discussed in that issue, but my inclination is that this is what we are actually looking for. Happy to discuss better ways if anyone can think of one, though.

### DIFF
--- a/src/Data/List/Lazy.purs
+++ b/src/Data/List/Lazy.purs
@@ -86,6 +86,7 @@ module Data.List.Lazy
   , transpose
 
   , foldM
+  , foldrLazy
 
   , module Exports
   ) where
@@ -716,3 +717,11 @@ foldM f a xs =
          Nothing -> pure a
          Just { head: b, tail: bs } ->
                        f a b >>= \a' -> foldM f a' bs
+
+-- | Perform a right fold lazily
+foldrLazy :: forall a b. Z.Lazy b => (a -> b -> b) -> b -> List a -> b
+foldrLazy op z = go
+  where
+    go xs = case step xs of
+      Cons x xs' -> Z.defer \_ -> x `op` go xs'
+      Nil -> z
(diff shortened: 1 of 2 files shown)


src/Data/List/Lazy.purs
```diff
@@ -705,3 +706,11 @@ foldM f a xs =
          Nothing -> pure a
          Just { head: b, tail: bs } ->
                        f a b >>= \a' -> foldM f a' bs
+
+-- | Perform a right fold lazily
+foldr' :: forall a b. Z.Lazy b => (a -> b -> b) -> b -> List a -> b
```
paf31: Can this be written with regular `foldr`?

matthewleon: Well, something like this will compile:

```purescript
foldr' op = foldr op'                                                                                                                                              
  where op' a b = Z.defer \_ -> a `op` b
```

But using `foldr` kills the laziness. So the test, for example, will just freeze on its infinite input.

Thinking if there is any kind of trick to be done here.

matthewleon: Honestly, given the definition of `foldr` as `foldl` (strict) on the reversed list, and the initial impetus for this change in #111 , I'm at a loss for how such a thing could be done.

Would love to know if you have something in mind.

paf31: I wasn't sure if it was possible, but thanks for trying. This looks good!

matthewleon: Ah, I thought you knew of a way and were attempting to deliver it from me Socratically :p

paf31:
Could you please rebase this? Thanks!

paf31:
Also can this be called `foldrLazy` instead? I'm not keen on using primes in general. Thanks!

matthewleon:
rebased and renamed

paf31:
Thanks!

PR #120, opened 2017-06-04
add stripPrefix and Pattern for Data.List

safareli:
https://github.com/purescript-contrib/purescript-parsing/pull/62#discussion_r119991248

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -482,11 +482,11 @@ sortBy cmp = mergeAll <<< sequences
 -- | A newtype used in cases where there is a list to be matched.
 newtype Pattern a = Pattern (List a)
 
-derive instance eqPattern :: (Eq a) => Eq (Pattern a)
-derive instance ordPattern :: (Ord a) => Ord (Pattern a)
+derive instance eqPattern :: Eq a => Eq (Pattern a)
+derive instance ordPattern :: Ord a => Ord (Pattern a)
 derive instance newtypePattern :: Newtype (Pattern a) _
 
-instance showPattern :: (Show a) => Show (Pattern a) where
+instance showPattern :: Show a => Show (Pattern a) where
   show (Pattern s) = "(Pattern " <> show s <> ")"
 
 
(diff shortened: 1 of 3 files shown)


safareli:
if you know a better place in the file for stripPrefix and Pattern let me know

src/Data/List.purs
```diff
@@ -476,6 +479,32 @@ sortBy cmp = mergeAll <<< sequences
 -- Sublists --------------------------------------------------------------------
 --------------------------------------------------------------------------------
 
+-- | A newtype used in cases where there is a list to be matched.
+newtype Pattern a = Pattern (List a)
+
+derive instance eqPattern :: (Eq a) => Eq (Pattern a)
+derive instance ordPattern :: (Ord a) => Ord (Pattern a)
+derive instance newtypePattern :: Newtype (Pattern a) _
```
safareli: not sure if this instance is correct

garyb: Yep, that's how deriving for `Newtype` works.

src/Data/List.purs
```diff
@@ -476,6 +479,32 @@ sortBy cmp = mergeAll <<< sequences
...
+-- | * `stripPrefix (Pattern (1:Nil)) (1:2:Nil) == Just (2:Nil)`
+-- | * `stripPrefix (Pattern Nil) (1:Nil) == Just (1:Nil)`
+-- | * `stripPrefix (Pattern (2:Nil)) (1:Nil) == Nothing`
+-- |
+-- | Running time: `O(n)` where `n` is the number of elements to strip.
+stripPrefix :: forall a. Eq a => Pattern a -> List a -> Maybe (List a)
+stripPrefix (Pattern p') s = tailRecM2 go p' s
+  where
+  go prefix input = case prefix, input of
+    (Cons p ps), (Cons i is) | p == i -> Just $ Loop ({ a: ps, b: is })
```
paf31: The parens are redundant in the patterns and in the value here.

src/Data/List.purs
```diff
@@ -476,6 +479,32 @@ sortBy cmp = mergeAll <<< sequences
 -- Sublists --------------------------------------------------------------------
 --------------------------------------------------------------------------------
 
+-- | A newtype used in cases where there is a list to be matched.
+newtype Pattern a = Pattern (List a)
+
+derive instance eqPattern :: (Eq a) => Eq (Pattern a)
```
paf31: Redundant parens.

paf31:
👍 Thanks!

paf31:
@garyb Do you want to use the `Pattern` newtype here too?

This looks good to me.

garyb:
The name `Pattern` is maybe a little strange in the context of lists, but I do think it's better to newtype with something so that the argument order is unambiguous. I can't think of a better name either, so 👍 I guess!

paf31:
`Prefix`?

garyb:
That works for sure, but if there's a chance any of the other string-ish functions will be replicated here then `Pattern` is more widely applicable. Even if just `stripSuffix` was added. :wink:

paf31:
True ... ok, merging.

PR #126, opened 2017-06-27
define stripPrefix for List.Lazy

safareli:
https://github.com/purescript-contrib/purescript-parsing/pull/62#discussion_r124173982


paf31:
Thanks!

PR #128, opened 2017-09-03
Require foldable-traversable ^3.3.0

MonoidMusician:
foldable-traversable@3.3.0 is the first release that includes `Traversable1` et al. that was added to this library in a recent release, so we need to make sure that at least that version is required. (None of the dependencies require more than ^3.0.0.)


paf31:
Thanks!

PR #130, opened 2017-09-07
Add dropEnd and takeEnd

notgiorgi:
Similar to the https://github.com/purescript/purescript-arrays/issues/113

### DIFF
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -60,8 +60,10 @@ module Data.List
   , stripPrefix
   , slice
   , take
+  , takeEnd
   , takeWhile
   , drop
+  , dropEnd
   , dropWhile
   , span
   , group
@@ -519,6 +521,13 @@ take = go Nil
   go acc _ Nil = reverse acc
   go acc n (x : xs) = go (x : acc) (n - 1) xs
 
+-- | Take the specified number of elements from the end of a list.
+-- |
+-- | Running time: `O(2n - m)` where `n` is the number of elements in list 
+-- | and `m` is number of elements to take.
+takeEnd :: forall a. Int -> List a -> List a
+takeEnd n xs = drop (length xs - n) xs
+
 -- | Take those elements from the front of a list which match a predicate.
 -- |
 -- | Running time (worst case): `O(n)`
@@ -535,6 +544,12 @@ drop :: forall a. Int -> List a -> List a
 drop 0 xs = xs
 drop _ Nil = Nil
 drop n (x : xs) = drop (n - 1) xs
+
+-- | Drop the specified number of elements from the end of a list.
+-- |
+-- | Running time: `O(2n - m)` where `n` is the number of elements to drop.
+dropEnd :: forall a. Int -> List a -> List a
+dropEnd n xs = take (length xs - n) xs
 
 -- | Drop those elements from the front of a list which match a predicate.
 -- |
--- a/src/Data/List.purs
+++ b/src/Data/List.purs
@@ -523,7 +523,7 @@ take = go Nil
 
 -- | Take the specified number of elements from the end of a list.
 -- |
--- | Running time: `O(2n - m)` where `n` is the number of elements in list 
+-- | Running time: `O(2n - m)` where `n` is the number of elements in list
 -- | and `m` is number of elements to take.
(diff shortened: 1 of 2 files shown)


test/Test/Data/List.purs
```diff
@@ -226,6 +226,11 @@ testList = do
   assert $ (take 2 (l [1, 2, 3])) == l [1, 2]
   assert $ (take 1 nil) == nil
 
+  log "take should keep the specified number of items from the end of an list, discarding the rest"
```
safareli: take**End** should keep the specified number of items from the end of an list, discarding the rest

test/Test/Data/List.purs
```diff
@@ -236,6 +241,11 @@ testList = do
   assert $ (drop 2 (l [1, 2, 3])) == l [3]
   assert $ (drop 1 nil) == nil
 
+  log "drop should remove the specified number of items from the front of an list"
```
safareli: drop**End** should remove the specified number of items from the **end** of an list

src/Data/List.purs
```diff
@@ -519,6 +521,13 @@ take = go Nil
   go acc _ Nil = reverse acc
   go acc n (x : xs) = go (x : acc) (n - 1) xs
 
+-- | Take the specified number of elements from the end of a list.
+-- |
+-- | Running time: `O(2n - m)` where `n` is the number of elements in list
+-- | and `m` is number of elements to take.
+takeEnd :: forall a. Int -> List a -> List a
+takeEnd n xs = drop (length xs - n) xs
```
paf31: This involves two traversals of the list, which we should try to avoid. Let's try to write these using `foldr` instead please.

notgiorgi: Is it even possible for this function to have better complexity? 

in order to take `n` elements from the end we need to start iterating the list from the start till we reach `length - n` position and return the tail. So we require length (whole list iteration) and we need to determine the index we stop on  (`length - n` elements iteration)

paf31: Yeah, ignore me, this is fine as it is. I thought we could do it in one pass with a buffer of size `n`, but it's probably not worth it. We can revisit later if necessary.

safareli: Alternative implementation will be to first reverse list and then take. it's performance will be n + m. it will be better then current version when m is smaller then n/2.

paf31: first reverse list and then take _and then reverse_

safareli: oh yeh :D so that will be better when m is less then n/4. so never mind

test/Test/Data/List.purs
```diff
@@ -226,16 +226,26 @@ testList = do
...
+  assert $ (takeEnd 2 (l [1, 2, 3])) == l [2, 3]
+  assert $ (takeEnd 1 nil) == nil
+
   log "takeWhile should keep all values that match a predicate from the front of an list"
   assert $ (takeWhile (_ /= 2) (l [1, 2, 3])) == l [1]
   assert $ (takeWhile (_ /= 3) (l [1, 2, 3])) == l [1, 2]
   assert $ (takeWhile (_ /= 1) nil) == nil
 
-  log "drop should remove the specified number of items from the front of an list"
+  log "dropEnd should remove the specified number of items from the end of an list"
```
safareli: You made this change in for wrong string :p

notgiorgi:
Looks good now?

test/Test/Data/List.purs
```diff
@@ -226,16 +226,26 @@ testList = do
...
+  assert $ (takeEnd 2 (l [1, 2, 3])) == l [2, 3]
+  assert $ (takeEnd 1 nil) == nil
+
   log "takeWhile should keep all values that match a predicate from the front of an list"
   assert $ (takeWhile (_ /= 2) (l [1, 2, 3])) == l [1]
   assert $ (takeWhile (_ /= 3) (l [1, 2, 3])) == l [1, 2]
   assert $ (takeWhile (_ /= 1) nil) == nil
 
-  log "drop should remove the specified number of items from the front of an list"
+  log "dropE should remove the specified number of items from the front of an list"
```
paf31: `drop` not `dropE` :)

But this is fine.

paf31:
Great, thanks!

PR #133, opened 2017-11-02
Avoid infinite loop possibility with takeEnd

mlms13:
Calling `takeEnd` with an `Int` value greater than the length of the list currently leads to an infinite loop. I added a couple test cases to demonstrate this and added a guard in `drop` against negative Ints.


paf31:
Thanks! Would you like to fix `take` too?

mlms13:
Good call, I added some tests and a guard there as well.

paf31:
Thanks!

PR #134, opened 2017-11-21
indexed foldable-traversable instances

matthewleon:


paf31:
Thanks!

Maybe we should deprecate `Data.List.mapWithIndex` now. Also, would you like to add `TraversableWithIndex` here too?

matthewleon:
Not sure how deprecation works in these libs, but I've done the rest.

paf31:
Thanks! I think a comment is enough for now.

paf31:
The alternative is to use a `Warn` constraint.

matthewleon:
Looks like I've made a test fail now. Will look into this.

matthewleon:
That test passes on my machine. Any idea what's going on there?

paf31:
It appears there is actually a bug:

```
> mapWithIndex (\x ix -> x + ix) (fromFoldable [0, 1, 2, 3])
(3 : 3 : 3 : 3 : Nil)
```

src/Data/List/Lazy/Types.purs
```diff
@@ -105,6 +107,12 @@ instance functorList :: Functor List where
     go Nil = Nil
     go (Cons x xs') = Cons (f x) (f <$> xs')
 
+instance functorWithIndexList :: FunctorWithIndex Int List where
+  mapWithIndex f xs = List (go 0 <$> unwrap xs)
+    where
+    go _ Nil = Nil
+    go i (Cons x xs') = Cons (f i x) (f (i + 1) <$> xs')
```
paf31: Shouldn't there be a recursive call to `go` here?

Maybe it's better to try to write this as a `foldr`?

matthewleon: Ah, indeed. I think I was running my tests locally on the wrong branch by accident.

matthewleon: Actually I think this is a different mistake I've made, besides the one breaking the test above. I'll go through and add some more thorough testing wherever it makes sense.

matthewleon:
Okay, this should be good to go now.

src/Data/List/Lazy/Types.purs
```diff
@@ -120,6 +125,22 @@ instance foldableList :: Foldable List where
 
   foldMap f = foldl (\b a -> b <> f a) mempty
 
+instance foldableWithIndexList :: FoldableWithIndex Int List where
+  foldrWithIndex f b xs =
```
paf31: This seems like we could extract a default implementation in terms of `foldlWithIndex`. What do you think?

matthewleon: I'll try.

matthewleon:
I don't think we can build a polymorphic default `foldrWithIndex` in terms of `foldlWithIndex`, actually. To begin, I don't think we can do a default `foldr` in terms of `foldl`.

My reasoning:

We know `foldr f a xs = foldl (flip f) a (reverse xs)`.

But this depends on a polymorphic `reverse`. 

There's one here for `Traversable`s: https://wiki.haskell.org/Foldable_and_Traversable
```haskell
reverseT :: (Traversable t) => t a -> t a
reverseT t = snd (mapAccumR (\ (x:xs) _ -> (xs, x)) (toList t) t)
```

But this, in turn, depends on a `toList`, and our library dependencies go in the other direction.

Now, we _do_ have a default implementation of `foldr` in terms of `foldMap`, which in turn is implemented in terms of `foldl`. But this uses the `Endo` monoid to build a thunk while traversing the `Foldable`, which kills our stack safety.

Please let me know if you think of something I've missed here. I would really love to get an efficient, stack-safe default implementation into `foldable-traversable`. But I can't think of any trick to do so.

paf31:
Looks good, and I think you're right about the default implementations.

Could you please add a test for each of the new methods though?

matthewleon:
Good idea. I will do that.

matthewleon:
tests done

src/Data/List/Types.purs
```diff
@@ -91,6 +110,11 @@ instance traversableList :: Traversable List where
   traverse f = map (foldl (flip (:)) Nil) <<< foldl (\acc -> lift2 (flip (:)) acc <<< f) (pure Nil)
   sequence = traverse id
 
+instance traversableWithIndexList :: TraversableWithIndex Int List where
+  traverseWithIndex f =
+    map (foldl (flip (:)) Nil)
```
paf31: Isn't this `reverse`?

test/Test/Data/List.purs
```diff
@@ -354,6 +375,13 @@ testList = do
   let xs = fromFoldable (range 1 100000)
   assert $ traverse Just xs == Just xs
 
+  log "traverseWithIndex should be stack-safe"
+  assert $ traverseWithIndex (const Just) xs == Just xs
+
+  log "traverseWithIndex should be correct"
+  assert $ traverseWithIndex (\i a -> Just $ i + a) (fromFoldable [2, 2, 2])
```
paf31: Nitpicking, sorry: could you use an Applicative which tests that the order of effects is correct, like `Array`?

paf31: Hmm, actually that wouldn't work, never mind.

src/Data/List/Types.purs
matthewleon: Yes, but we can't use it because of the reverse dependency.

paf31:
Thanks!

Issue #131, opened 2017-09-22
A new List.map that is both stack-safe and fast


paulyoung:
I came across this and thought it was worth sharing: https://discuss.ocaml.org/t/a-new-list-map-that-is-both-stack-safe-and-fast/865

hdgarrood:
Fixed by #135 / #157.

PR #135, opened 2017-11-21
List Functor: mix unrolled and reverse map

matthewleon:
Addresses #131

The relevant chunk sizes (5 for the initial list segment), (3 for the
tail-recursive remainder) were arrived at through benchmarked
experimentation, mapping a simple (_ + 1) through lists of various
sizes.

Relevant figures:
list of 1000 elems:   142.61 μs -> 36.97 μs
list of 2000 elems:   275.17 μs -> 55.33 μs
list of 10000 elems:  912.73 μs -> 208.39 μs
list of 100000 elems: 34.56 ms  -> 1.24 ms

The ~30x speed increase for long lists is probably explained by the lack
of GC thrashing with this approach.

### DIFF
--- a/src/Data/List/Types.purs
+++ b/src/Data/List/Types.purs
@@ -79,33 +79,18 @@ instance functorList :: Functor List where
 -- https://discuss.ocaml.org/t/a-new-list-map-that-is-both-stack-safe-and-fast/865
 -- chunk sizes determined through experimentation
 listMap :: forall a b. (a -> b) -> List a -> List b
-listMap f = startUnrolledMap unrollLimit where
-  -- iterate the unrolled map up to 200 times,
-  -- which hits up to 1000 elements
-  unrollLimit = 200
-
-  startUnrolledMap :: Int -> List a -> List b
-  startUnrolledMap 0 (x : xs) = f x : chunkedRevMap xs
-  startUnrolledMap n (x1 : x2 : x3 : x4 : x5 : xs) =
-    f x1 : f x2 : f x3 : f x4 : f x5 : startUnrolledMap (n - 1) xs
-  startUnrolledMap n (x1 : x2 : x3 : x4 : xs) =
-    f x1 : f x2 : f x3 : f x4 : startUnrolledMap (n - 1) xs
-  startUnrolledMap n (x1 : x2 : x3 : xs) =
-    f x1 : f x2 : f x3 : startUnrolledMap (n - 1) xs
-  startUnrolledMap n (x1 : x2 : xs) =
-    f x1 : f x2 : startUnrolledMap (n - 1) xs
-  startUnrolledMap n (x : xs) = f x : startUnrolledMap (n - 1) xs
-
-  startUnrolledMap _ Nil = Nil
-
-  chunkedRevMap :: List a -> List b
-  chunkedRevMap = go Nil
+listMap f = chunkedRevMap Nil
+  where
+  chunkedRevMap :: List (List a) -> List a -> List b
+  chunkedRevMap chunksAcc chunk@(x1 : x2 : x3 : xs) =
+    chunkedRevMap (chunk : chunksAcc) xs
+  chunkedRevMap chunksAcc xs =
+    reverseUnrolledMap chunksAcc $ unrolledMap xs
     where
-    go :: List (List a) -> List a -> List b
-    go chunksAcc chunk@(x1 : x2 : x3 : xs) =
-      go (chunk : chunksAcc) xs
-    go chunksAcc finalChunk =
-      reverseUnrolledMap chunksAcc $ startUnrolledMap 0 finalChunk
+    unrolledMap :: List a -> List b
+    unrolledMap (x1 : x2 : Nil) = f x1 : f x2 : Nil
+    unrolledMap (x1 : Nil) = f x1 : Nil
+    unrolledMap _ = Nil
 
     reverseUnrolledMap :: List (List a) -> List b -> List b
     reverseUnrolledMap ((x1 : x2 : x3 : _) : cs) acc =
--- a/src/Data/List/Types.purs
+++ b/src/Data/List/Types.purs
@@ -2,6 +2,8 @@ module Data.List.Types
   ( List(..)
   , (:)
   , NonEmptyList(..)
+  , toList
+  , nelCons
   ) where
 
 import Prelude
--- a/src/Data/List/Types.purs
+++ b/src/Data/List/Types.purs
@@ -102,7 +102,7 @@ listMap f = startUnrolledMap unrollLimit where
   chunkedRevMap = go Nil
     where
     go :: List (List a) -> List a -> List b
-    go chunksAcc chunk@(x1 : x2 : x3 : x4 : x5 : xs) =
+    go chunksAcc chunk@(x1 : x2 : x3 : xs) =
       go (chunk : chunksAcc) xs
     go chunksAcc finalChunk =
       reverseUnrolledMap chunksAcc $ startUnrolledMap 0 finalChunk
--- a/src/Data/List/Types.purs
+++ b/src/Data/List/Types.purs
@@ -1,4 +1,8 @@
-module Data.List.Types where
+module Data.List.Types
+  ( List(..)
+  , (:)
+  , NonEmptyList(..)
+  ) where
 
 import Prelude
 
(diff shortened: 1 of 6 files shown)


matthewleon:
Will update this with details on my machine setup.

src/Data/List/Types.purs
```diff
@@ -65,7 +64,47 @@ instance monoidList :: Monoid (List a) where
   mempty = Nil
 
 instance functorList :: Functor List where
-  map f = foldr (\x acc -> f x : acc) Nil
+  map = listMap
+
+-- chunked list Functor inspired by OCaml
+-- https://discuss.ocaml.org/t/a-new-list-map-that-is-both-stack-safe-and-fast/865
+-- chunk sizes determined through experimentation
+listMap :: forall a b. (a -> b) -> List a -> List b
```
paf31: This is exported right now, is that deliberate?

src/Data/List/Types.purs
```diff
@@ -65,7 +64,47 @@ instance monoidList :: Monoid (List a) where
...
+listMap f = startUnrolledMap unrollLimit
+  where
+  -- iterate the unrolled map up to 1000 times,
+  -- which hits up to 5000 elements
+  unrollLimit = 1000
+
+  startUnrolledMap :: Int -> List a -> List b
+  startUnrolledMap 0 (x : xs) = f x : chunkedRevMap xs
+  startUnrolledMap n (x1 : x2 : x3 : x4 : x5 : xs) =
+    f x1 : f x2 : f x3 : f x4 : f x5 : startUnrolledMap (n - 1) xs
```
paf31: I'm not sure if OCaml has tail calls modulo cons, but isn't this going to lead to a stack overflow for us?

src/Data/List/Types.purs
```diff
@@ -65,7 +64,47 @@ instance monoidList :: Monoid (List a) where
...
+listMap :: forall a b. (a -> b) -> List a -> List b
+listMap f = startUnrolledMap unrollLimit
+  where
+  -- iterate the unrolled map up to 1000 times,
+  -- which hits up to 5000 elements
+  unrollLimit = 1000
+
+  startUnrolledMap :: Int -> List a -> List b
+  startUnrolledMap 0 (x : xs) = f x : chunkedRevMap xs
+  startUnrolledMap n (x1 : x2 : x3 : x4 : x5 : xs) =
```
paf31: This can probably be made faster if you hand-optimize the cases. The JS generated here isn't going to be particularly optimal. We can reuse the branches.

paf31:
Thanks!

Another thing to try is a straight traversal of the spine using a mutable cons cell, a la Scheme. I think we could make this safe. What do you think?

src/Data/List/Types.purs
matthewleon: No. It isn't. Good catch.

src/Data/List/Types.purs
matthewleon: Yes, I thought the same thing. I'll look at it if the option you suggest with mutation isn't the way to go.

src/Data/List/Types.purs
matthewleon: This is why I've limited the "unrolled map" to 1000 iterations. It should avoid stack overflow. Are you suggesting that there is a way to make it stack overflow despite that limit?

matthewleon:
> straight traversal of the spine using a mutable cons cell, a la Scheme

I will research this and see if I can come up with something crafty. Is the idea here that we jump into `ST`?

paf31:
I think it would be FFI instead of ST unfortunately, unless we can come up with a way to modify cons cells in `ST`, which would mean some sort of weird `STList` data type.

src/Data/List/Types.purs
paf31: I guess not, unless `f` was already on the verge of overflowing the stack. It can only make an existing problem a finite amount worse :)

matthewleon:
> unless we can come up with a way to modify cons cells in ST, which would mean some sort of weird STList data type.

I was thinking along those lines. I'll try it and benchmark it.

matthewleon:
Though upon further thought, I don't see how to do that without running into the problem of stack-safe recursion on `Eff` being very slow. Perhaps the saner thing to do is to try the mutable cons cell in FFI... I'll have to use JS anyway to optimize the unroll.

matthewleon:
Further, outside of `ST`, I'm having trouble figuring out how to write JavaScript for this that isn't dependent on the way the compiler chooses to structure the `List` type, which seems very brittle.

src/Data/List/Types.purs
matthewleon: Doesn't hand-optimizing mean that the code could break whenever the compiler chooses to lay out data differently?

paf31: No, I just mean rewriting this so that common cases use the same branch, instead of having multiple branches which recognize lists of length >= 1 for example.

paf31: If you look at the generated JS, you'll hopefully see what I mean.

natefaubion:
```purescript
case xs of
  x : xs' -> ...
    case xs' of
      x' : xs'' -> ...
      _ -> ...
  _ -> ...
```

PureScript doesn't perform this optimization, you have to do it manually.

paf31:
Yes, agreed, it doesn't seem worth it to make the code non-portable.

matthewleon:
> Yes, agreed, it doesn't seem worth it to make the code non-portable.

So do you prefer that the code be largely left as-is, because this version is portable and has much better performance than what we had before, at least in Node?

Or should I go ahead and try to do the mutable cons cell + unrolling in FFI JavaScript?

The answer to me really isn't obvious, as I'm unaware of what PS's stance regarding library portability is. As it stands, this lib doesn't use any FFI at all – directly. I believe that `Lazy`, however, uses FFI. Which makes at least the lazy lists non-portable, no?

paf31:
Well the lazy data structure doesn't rely on the compiler internals in any way.

Here is a quick FFI experiment:

```javascript
exports.mapList = function(f) {
  return function(l) {
    if (l.constructor.name === 'Nil') {
      return l;
    } else {
      var nil = require('Data.List.Types').Nil.value;
      var cons = require('Data.List.Types').Cons;
      var result = new cons(null, null);

      var input = l;
      var output = result;

      while (true) {
        output.value0 = f(input.value0);
        output.value1 = new cons(null, null);
        
        if (input.value1.constructor.name == 'Nil') {
          break;
        }

        input = input.value1;
        output = output.value1;
      }

      output.value1 = nil;

      return result;
    }
  }
}
```

It gets about 3ms on a list of 100000 elements on my machine, but I'd be interested to hear what you get.

So it seems like the loop unrolling is really helping quite a bit here.

matthewleon:
Recapping conversation with @paf31 on Slack: Though considerable speed gains can be made by using FFI with a mutable Cons cell, the List library is currently portable, and we don't want to introduce FFI just to optimize `map` further than this.

paf31:
I've put my code in a repo [here](https://github.com/paf31/purescript-lists-fast), along with `zipWith` and `filter` replacements.

matthewleon:
Cool. To be sure, those expected speedups might decrease a fair bit once we
merge in non-FFI optimizations to Data.List.

matthewleon:
I've added benchmarking to this branch.

safareli:
`package-lock.json`  was committed in last commit

matthewleon:
@safareli removed it.

src/Data/List/Types.purs
```diff
@@ -65,7 +68,47 @@ instance monoidList :: Monoid (List a) where
...
 instance functorList :: Functor List where
-  map f = foldr (\x acc -> f x : acc) Nil
+  map = listMap
+
+-- chunked list Functor inspired by OCaml
+-- https://discuss.ocaml.org/t/a-new-list-map-that-is-both-stack-safe-and-fast/865
+-- chunk sizes determined through experimentation
+listMap :: forall a b. (a -> b) -> List a -> List b
+listMap f = startUnrolledMap unrollLimit
+  where
```
paf31: Nitpick: could you please indent things past the `where`, or move the `where` up to the end of the previous line?

paf31:
Thanks!

What do other people think? @garyb @hdgarrood 

Especially about the potential stack overflow issue mentioned above.

garyb:
@paf31 what was the stack overflow issue? Just that thing that if the stack is near-overflow already it might push it over the edge?

hdgarrood:
Yeah, if `f` makes enough recursive calls on average, I think that could cause stack overflows. This looks really good though! If you could address @paf31's recent comments and do a little testing to check that stack-safety isn't significantly worse than it is now, I'm :+1: to merge.

Come to think of it, do we have any testing for stack overflows in `map` in this library already?

paf31:
To be clear, the issue is that we go from "`map` never grows the stack" to "`map` always grows the stack, but maybe not by very much". The growth should be linear in the size of the list.

natefaubion:
With this, isn't it "map grows the stack by a fixed maximum amount"?

paf31:
Well `startUnrolledMap` calls itself recursively, but not tail-recursively, so no, I think the stack usage grows with the size of the input.

natefaubion:
Right, but it has a maximum threshold defined be `unrollLimit`, which causes it to fallback to the slower, stack-safe variant which halts the growth. So at most, `map` itself will consume 1000 frames .

paf31:
Oh, of course, sorry.

safareli:
If frame limit is 10,000 and before map is called 9,001 frames are used then using 1k frames will overflow the stack right? is this fine?

paf31:
Nested `map`s are also quite common. I don't think this is enough to cause a problem on its own, but it could worsen it. The question for me is not whether this code should be included, but whether it should be the default. Do we want a default which works in every case but is slow in most cases, or one which works in most cases but overflows the stack in a tiny number of cases?

I think it's probably okay to pick this as the default, but I just want to make sure the trade off is clear now as we decide.

natefaubion:
There's also the question of what the stack limit is. It does 5 items at a time, so it will eat stack up to 5000 items. Maybe we don't need to handle that many, and we can reduce it to 200 frames if it's a concern?

matthewleon:
> There's also the question of what the stack limit is. It does 5 items at a time, so it will eat stack up to 5000 items. Maybe we don't need to handle that many, and we can reduce it to 200 frames if it's a concern?

I wouldn't mind doing that. Maybe it's worth me doing some exploratory tests to see how much breathing room Node gives us?

matthewleon:
regarding the call stack size: http://2ality.com/2014/04/call-stack-size.html

matthewleon:
> we can reduce it to 200 frames if it's a concern?

I've made a commit that does this. I figure that this means we still get a big speed-up for common List use-cases, while if people are serious about having faster iteration through longer lists, they can use the fast list repo.

matthewleon:
rebased.

paf31:
Given the possible stack-safety issue, what do people think about providing the naive (safe) implementation as a separate named function? Or vice versa?

matthewleon:
One possibility is to just use the stack-safe unrolled operation, which should still give a speed boost, and recommend that people turn to `purescript-lists-fast` if speed is important.

paf31:
Would it be possible to try that and compare the performance easily?

matthewleon:
I made a branch that only uses the stack-safe reverse map and... It's actually faster.

map: list (2000 elems)
mean   = 53.33 μs -> 31.24 μs

map: list (10000 elems)
mean   = 107.73 μs -> 98.17 μs

I've merged it in.

matthewleon:
Just realized I can also inline the `unrolledMap`, will do that and push.

matthewleon:
Actually, inlining `unrolledMap` slows things down significantly. Mysteries of JS... This should be good for review.

paf31:
Even better, thanks!

Any other comments before I merge this?

hdgarrood:
Merged in via #157.

Issue #138, opened 2018-01-03
Bug of Data.List.Lazy take


ibara1454:
My English isn't good so feel free to ask me if there is anything unclear.

## Problem

I found a problem that `take` evaluate the value too early.
It always evaluate the **next** value which we don't need.

To show this, I construct a lazy constructed list (elements in the list will be constructed only when they are needed), and use `spy` to print the value when the it is evaluated.

## Actually

![bug_take](https://user-images.githubusercontent.com/4210652/34532339-2e9e2558-f0f9-11e7-9bf7-b9e75c362ba3.JPG)

## Desire

```purescript
> take 0 lazyl
fromStrict (Nil)

> take 1 lazyl
1
fromStrict ((Cons 1 Nil))
```

ibara1454:
And I have made a pull request of this.
You can use it if you think it is OK.

PR #139, opened 2018-01-03
Don't force more than necessary in Data.List.Lazy take

ibara1454:
#138 

Made a fix of `take`.

fix_take

### DIFF
--- a/src/Data/List/Lazy.purs
+++ b/src/Data/List/Lazy.purs
@@ -509,8 +509,9 @@ slice start end xs = take (end - start) (drop start xs)
 -- |
 -- | Running time: `O(n)` where `n` is the number of elements to take.
 take :: forall a. Int -> List a -> List a
-take n _ | n <= 0 = nil
-take n xs = List <<< map (go n) <<< unwrap $ xs
+take n = if n <= 0
+  then const nil
+  else List <<< map (go n) <<< unwrap
   where
   go :: Int -> Step a -> Step a
   go _ Nil = Nil
--- a/src/Data/List/Lazy.purs
+++ b/src/Data/List/Lazy.purs
@@ -509,10 +509,10 @@ slice start end xs = take (end - start) (drop start xs)
 -- |
 -- | Running time: `O(n)` where `n` is the number of elements to take.
 take :: forall a. Int -> List a -> List a
-take n = List <<< map (go n) <<< unwrap
+take n _ | n <= 0 = nil
+take n xs = List <<< map (go n) <<< unwrap $ xs
   where
   go :: Int -> Step a -> Step a
-  go i _ | i <= 0 = Nil
   go _ Nil = Nil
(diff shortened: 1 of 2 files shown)


src/Data/List/Lazy.purs
```diff
@@ -509,10 +509,10 @@ slice start end xs = take (end - start) (drop start xs)
 -- |
 -- | Running time: `O(n)` where `n` is the number of elements to take.
 take :: forall a. Int -> List a -> List a
-take n = List <<< map (go n) <<< unwrap
+take n _ | n <= 0 = nil
+take n xs = List <<< map (go n) <<< unwrap $ xs
```
matthewleon: Tiny nitpick: You could probably just do `$ unwrap xs`, no?

matthewleon: (Or maintain the point-free form used previously.)

matthewleon:
This looks good to me.

hdgarrood:
:+1: for keeping the previous pointfree form. Also would you mind adding a test? Perhaps something like this?

```purescript
> oops x = oops x
> xs = defer \_ -> 1 : defer oops
> take 1 xs == fromFoldable [1]
```

The idea being this will test that we don't evaluate any more of the list than needed, since if we tried to, it would crash with a stack overflow.

ibara1454:
Thanks for the replys.:smiley:
About the style, I have tried the pointfree form.
Like this

``` purescript
take :: forall a. Int -> List a -> List a
take n | n <= 0 = const nil
take n = List <<< map (go n) <<< unwrap
```

But I think it is a little difficult to understand, since it used an additional `const`.
Is it better than the current one?

hdgarrood:
Yes, although come to think of it I think using `if n <= 0 then const nil else ...` would be better style here since we’re matching on the arguments in the same way in both alternatives currently.

ibara1454:
Done.:smiley:
- Rewrite in pointfree style
- Add test case

(I use `oops x = 0 : (oops x)` instead, since `oops x = oops x` is tail recursive)

test/Test/Data/List/Lazy.purs
```diff
@@ -270,6 +270,11 @@ testListLazy = do
   assert $ (take 2 (l [1, 2, 3])) == l [1, 2]
   assert $ (take 1 nil') == nil'
 
+  log "take should evaluate exactly n items which we needed"
+  assert let oops x = 0 : (oops x)
```
hdgarrood: sorry to nitpick here, but the parens around `oops x` aren't necessary and I think it would be better style to remove them, and also I think it would be a little nicer (and more in line with the style used elsewhere in this file) to use parens instead of `$` here too.

ibara1454: So I see! parens are not unnecessary here. Thanks for your correction.

ibara1454:
Fixed.

- Remove unnecessary parens
- Remove `($)`

Issue #140, opened 2018-01-03
Bug of Data.List.Lazy iterate


ibara1454:
My English isn't good so feel free to ask me if there is anything unclear.

## Problem

The correct behavior of `iterate`, it should evaluate the function only `n` times while generating a `n+1` length list. But in this implementation, `iterate` re-evaluate values too many times.

To show this, I constructed a list by `iterate`, and the value in list will be printed to console when it is evaluated.

## Actually

![bug_iterate](https://user-images.githubusercontent.com/4210652/34533727-d46b2a76-f0fe-11e7-8cd4-004dd52190bf.JPG)

## Desire

```purescript
> lazyl = iterate (\x -> spy $ x + 1) 0
> take 3 lazyl
1
2
fromStrict ((Cons 0 (Cons 1 (Cons 2 Nil))))

> take 4 lazyl
1
2
3
fromStrict ((Cons 0 (Cons 1 (Cons 2 (Cons 3 Nil)))))
```

ibara1454:
Made a PR #141 for this issue.

However, maybe there still have some bugs in `fix` or `(<$>)`. Because the (old) code

```purescript
iterate f x = Z.fix \xs -> cons x (f <$> xs)
```

should not have the behavior that evaluate the same value too many times.

natefaubion:
@ibara1454 My hunch would be that the `Lazy` instance does not actually share work.

ibara1454:
@natefaubion OK. To confirm it, I rewrote `fix` without `Lazy` class.

```purescript
import Prelude
import Data.List.Lazy (List(..), iterate, cons, step)
import Control.Lazy (fix)
import Data.Lazy (defer)
import Debug.Trace (spy)

defer' :: forall a. (Unit -> List a) -> List a
defer' f = List $ defer (step <<< f)

fix' :: forall a. (List a -> List a) -> List a
fix' f = defer' (\_ -> f (fix' f))

iterate' :: forall a. (a -> a) -> a -> List a
iterate' f x = fix' \xs -> cons x (f <$> xs)

ls = iterate (\x -> spy $ x + 1) 0
ls' = iterate' (\x -> spy $ x + 1) 0
```

But they have the same result ...
![compare_fix](https://user-images.githubusercontent.com/4210652/34646622-608320c8-f3b0-11e7-8f1c-60573a9f1c32.JPG)
It seems that the bug is made by `(<$>)`, or by `defer`, or by `fix` itself.

[...]

PR #141, opened 2018-01-03
Fix Data.List.Lazy iterate

ibara1454:
A solution of #140 

Here's the result of the modification of `iterate`.
All the values are evaluated exactly once. 

fix_iterate

It still has a bug. Because the result of `take 3 lazyl` should be

```purescript
> take 3 lazyl
1
2
fromStrict ((Cons 0 (Cons 1 (Cons 2 Nil))))
```

Not

```purescript
> take 3 lazyl
1
2
3
fromStrict ((Cons 0 (Cons 1 (Cons 2 Nil))))
```

But the bug is made by `take` (as #138), not `iterate`. So we have nothing to do with this.

Just to be sure, I made a performance comparison between the before and after of `iterate`.
It is made by compiling the purescript code into javascript, and then run the benchmark by Benchmark.js.

Each value in this benchmark is the time of the calculation `sum $ take n $ iterate (_ + 1) 1`, where `n` is the length of list in the benchmark.

The old `iterate` has no value after length > 5000, because it throws exceptions by maximum call stack size exceeded. Yes, it is not stack-safe.

iterate

The old `iterate` run extremely slow in PSCi (time complexity is not linear), but (I don't know why) it is not bad (except the stack overflow) in this benchmark.

### DIFF
--- a/src/Data/List/Lazy.purs
+++ b/src/Data/List/Lazy.purs
@@ -186,7 +186,7 @@ repeat x = Z.fix \xs -> cons x xs
 
 -- | Create a list by iterating a function
 iterate :: forall a. (a -> a) -> a -> List a
-iterate f x = Z.fix \xs -> cons x (f <$> xs)
+iterate f x = cons x (Z.defer \_ -> iterate f (f x))
 
 -- | Create a list by repeating another list
 cycle :: forall a. List a -> List a


ibara1454:
The bug is based on `fix`.

PR #144, opened 2018-01-27
NonEmptyList: newtype derive Foldable1

matthewleon:
Depends on https://github.com/purescript/purescript-nonempty/pull/27

Will update `bower.json` if that gets merged.


matthewleon:
Updated as promised.

PR #150, opened 2018-07-31
fix alt for ZipList

xgrommx:


garyb:
Seems reasonable to me. Anyone else?

src/Data/List/ZipList.purs
```diff
@@ -46,7 +47,7 @@ instance applicativeZipList :: Applicative ZipList where
   pure = ZipList <<< repeat
 
 instance altZipList :: Alt ZipList where
-  alt = append
+  alt (ZipList xs) (ZipList ys) = ZipList $ (xs <> drop (length xs) ys)
```
natefaubion: Unnecessary `$`, or unnecessary parens.

hdgarrood:
Can you describe what is being fixed here? Also, do you have a proof that this version satisfies all the relevant laws?

hdgarrood:
Also won’t this diverge if the first argument is infinite?

hdgarrood:
Ah ok, so I presume this is motivated by the fact that the current version doesn’t satisfy the Alternative distributivity law
> (f <|> g) <*> x == (f <*> x) <|> (g <*> x)
since with the current instance, the left hand side will be a list of length `min(length f + length g, x)`, whereas the right hand side will be a list of length `min(length f, length x) + min(length g, length x)`. For example, with `f = g = ZipList ((_+1) : nil)` and `x = ZipList (0 : nil)` we get the left hand side being `ZipList (1 : nil)` but the right hand side being `ZipList (1 : 1 : nil)`.

thomashoneyman (approved):

hdgarrood (approved):

hdgarrood:
I'm going to merge this momentarily, just as soon as I've written a satisfactory commit message

PR #155, opened 2019-02-06
Add scanrLazy for lazy lists

drewolson:


garyb:
Thanks!

drewolson:
Thanks for merging @garyb. I was unsure if I should be exposing a new function named `scanrLazy` or be removing the re-exported `scanr` from `Data.Traversable` and implementing this as `scanr`.

The file seems inconsistent (e.g. `filter` vs. `foldrLazy`). Thoughts?

garyb:
Yeah, I think it's a bit of a mess in there at the moment, perhaps next time we make a breaking release we can clean things up a bit!

PR #157, opened 2019-03-10
fast functor updated

hdgarrood:
This is @matthewleon's faster `List.map`, from #135, updated for 0.12.

### DIFF
--- a/src/Data/List/Types.purs
+++ b/src/Data/List/Types.purs
@@ -95,9 +95,7 @@ listMap f = chunkedRevMap Nil
     reverseUnrolledMap :: List (List a) -> List b -> List b
     reverseUnrolledMap ((x1 : x2 : x3 : _) : cs) acc =
       reverseUnrolledMap cs (f x1 : f x2 : f x3 : acc)
-    -- if we pattern match on Nil, we need a Partial constraint,
-    -- which kills TCO
-    reverseUnrolledMap _ acc = acc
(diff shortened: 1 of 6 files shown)


src/Data/List/Types.purs
```diff
@@ -67,7 +73,31 @@ instance monoidList :: Monoid (List a) where
...
+    unrolledMap :: List a -> List b
+    unrolledMap (x1 : x2 : Nil) = f x1 : f x2 : Nil
+    unrolledMap (x1 : Nil) = f x1 : Nil
+    unrolledMap _ = Nil
+
+    reverseUnrolledMap :: List (List a) -> List b -> List b
+    reverseUnrolledMap ((x1 : x2 : x3 : _) : cs) acc =
+      reverseUnrolledMap cs (f x1 : f x2 : f x3 : acc)
+    -- if we pattern match on Nil, we need a Partial constraint,
+    -- which kills TCO
```
natefaubion: FWIW I don't think partial constraints kill tco anymore.

hdgarrood: Good point. Shall we just remove this comment then?

natefaubion: 👍

matthewleon:
Cool to see this happen!

hdgarrood:
@matthewleon thanks again for this, and sorry it slipped under the radar for so long!

PR #169, opened 2020-04-23
Fix lazy list docs where original is returned instead of nothing

milesfrain:


src/Data/List/Lazy.purs
```diff
@@ -326,11 +326,8 @@ findIndex fn = go 0
 findLastIndex :: forall a. (a -> Boolean) -> List a -> Maybe Int
 findLastIndex fn xs = ((length xs - 1) - _) <$> findIndex fn (reverse xs)
 
--- | Insert an element into a list at the specified index, returning a new
--- | list or `Nothing` if the index is out-of-bounds.
--- |
--- | This function differs from the strict equivalent in that out-of-bounds arguments
--- | result in the element being appended at the _end_ of the list.
+-- | Insert an element into a list at the specified index, or append the element
+-- | to the _end_ of the list if the index is out-of-bounds, returning a new list.
```
hdgarrood, line 330: I don't think the italics are necessary here

milesfrain: Sure. Was just preserving original text.
Updated PR.

hdgarrood: Oh oops sorry, I thought that was new. Thanks anyway!

PR #177, opened 2020-12-03
Migrate to GitHub Actions

thomashoneyman:
This PR migrates lists to use GitHub Actions instead of Travis for CI, as discussed in [semirings#17](https://github.com/purescript/purescript-semirings/pull/17). It also updates the installation instructions to refer to [Spago](https://github.com/purescript/spago) instead of Bower.


thomashoneyman:
I've removed the `MonadZero` instances in this library as the class was [deprecated as discussed in  control#62](https://github.com/purescript/purescript-control/issues/62). However, I'm not totally clear on whether we are:

* Preserving `MonadZero` and `MonadZero` instances across core until the class is removed, where `control` and any other library using the class will have a compiler warning for the deprecation, or
* Preserving `MonadZero` only in `control`, and removing it from the other libraries in core.

Either way, we won't be able to use the `censor-lib --strict` command with `psa` in CI anymore for at least any library which is using the class, because we _want_ this warning to be included in the passing build. See this failed build for what happens in CI right now with the warning:

https://github.com/purescript/purescript-lists/runs/1490231947

thomashoneyman:
I've re-added the instances to comply with the discussion in https://github.com/purescript/purescript-control/issues/62. We'll have to take an additional step to get CI working with this warning.

thomashoneyman:
Backlinking to https://github.com/purescript/purescript/issues/3962

thomashoneyman:
Tests pass due to censoring user-defined warnings. Backlinking to https://github.com/purescript/purescript-lists/issues/176.

hdgarrood (approved):
Thanks!

PR #179, opened 2020-12-18
Match nub* functions with Array

### LINKED ISSUE #178: Match nub* functions with Array
The names and signatures of the `nub*` functions for `List` and `Array` are inconsistent.

```purs
-- List
nub :: forall a. Eq a => List a -> List a
nubBy :: forall a. (a -> a -> Boolean) -> List a -> List a

-- Array
nub :: forall a. Ord a => Array a -> Array a
nubEq :: forall a. Eq a => Array a -> Array a
nubBy :: forall a. (a -> a -> Ordering) -> Array a -> Array a
nubByEq :: forall a. (a -> a -> Boolean) -> Array a -> Array a
```

Proposing we make List match Array and use this lineup:
```purs
nub :: forall a. Ord a => List a -> List a
nubEq :: forall a. Eq a => List a -> List a
nubBy :: forall a. (a -> a -> Ordering) -> List a -> List a
nubByEq :: forall a. (a -> a -> Boolean) -> List a -> List a
```

We should also include examples and runtime info.

milesfrain:
Fixes #178

- [x] `Data.List`
- [x] `Data.List.Lazy`
- [x] `Data.List.NonEmpty`

Edit: Now with `nub`/`nubBy` for lazy list too.

-----------
Old comment:

Since `Data.List.Lazy` doesn't have `sort` yet, I don't think it's in the scope of this PR to add the optimized `Ord`-based `nub`/`nubBy` to that module. It will just be stuck with `nubEq`/`nubByEq` for now.

If we do want to add `sort` to `Data.List.Lazy` (in another PR), we might be able to find some inspiration from [Haskell's implementation](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-List.html#g:21).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,310 @@
+# Changelog
+
+Notable changes to this project are documented in this file. The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+Breaking changes:
+
+New features:
+
+Bugfixes:
+
+Other improvements:
+
+## [v5.4.1](https://github.com/purescript/purescript-lists/releases/tag/v5.4.1) - 2019-05-06
+
+Fixes a shadowed name warning in advance of new `purs` version
+
+## [v5.4.0](https://github.com/purescript/purescript-lists/releases/tag/v5.4.0) - 2019-02-09
+
+Added `scanrLazy` for lazy lists (@drewolson)
+
+## [v5.3.0](https://github.com/purescript/purescript-lists/releases/tag/v5.3.0) - 2018-08-31
+
+- Added `cons'` and `snoc'` variations for `NonEmptyList` that construct from an `a`+`List a`
+
+## [v5.2.0](https://github.com/purescript/purescript-lists/releases/tag/v5.2.0) - 2018-08-19
+
+- Added `Unfoldable1` instances for `NonEmptyList`s
+
+## [v5.1.0](https://github.com/purescript/purescript-lists/releases/tag/v5.1.0) - 2018-08-19
+
+- Added `FunctorWithIndex` and related instances for `NonEmptyList`s (@cryogenian)
+
+## [v5.0.0](https://github.com/purescript/purescript-lists/releases/tag/v5.0.0) - 2018-05-23
+
+- Updated for PureScript 0.12
+- Fixed a bug in `take` for lazy lists (@ibara1454)
+- Added `tails` (@colehaus)
+
+## [v4.12.0](https://github.com/purescript/purescript-lists/releases/tag/v4.12.0) - 2017-11-30
+
+Add indexed `Functor`, `Foldable` and `Traversable` instances (@matthewleon)
+
+## [v4.11.0](https://github.com/purescript/purescript-lists/releases/tag/v4.11.0) - 2017-11-02
+
+Short circuit `drop` and `take` on negative inputs (@mlms13)
+
+## [v4.10.0](https://github.com/purescript/purescript-lists/releases/tag/v4.10.0) - 2017-09-09
+
+Add `dropEnd` and `takeEnd` (@notgiorgi)
+
+## [v4.9.1](https://github.com/purescript/purescript-lists/releases/tag/v4.9.1) - 2017-09-03
+
+Require `foldable-traversable` explicitly (@MonoidMusician)
+
+## [v4.9.0](https://github.com/purescript/purescript-lists/releases/tag/v4.9.0) - 2017-07-06
+
+Add `stripPrefix` for `List.Lazy` (@safareli)
+
+## [v4.8.0](https://github.com/purescript/purescript-lists/releases/tag/v4.8.0) - 2017-06-30
+
+- Added most list operations for `NonEmptyList`
+- Added `Foldable1` and `Traversable1` instances for `NonEmptyList`
+
+## [v4.7.0](https://github.com/purescript/purescript-lists/releases/tag/v4.7.0) - 2017-06-13
+
+- Added `cons`, `snoc`, and `unsnoc` for `Data.List.NonEmpty` (@natefaubion)
+
+## [v4.6.1](https://github.com/purescript/purescript-lists/releases/tag/v4.6.1) - 2017-06-07
+
+Require `tailrec^3.3.0`
+
+## [v4.6.0](https://github.com/purescript/purescript-lists/releases/tag/v4.6.0) - 2017-06-04
+
+Add `stripPrefix` and `Pattern` to `Data.List` (@safareli)
+
+## [v4.5.0](https://github.com/purescript/purescript-lists/releases/tag/v4.5.0) - 2017-06-04
+
+Add `foldrLazy` (@matthewleon)
+
+## [v4.4.0](https://github.com/purescript/purescript-lists/releases/tag/v4.4.0) - 2017-06-03
+
+Add `partition` (@matthewleon)
+
+## [v4.3.0](https://github.com/purescript/purescript-lists/releases/tag/v4.3.0) - 2017-05-28
+
+- `sort` and `sortBy` for `NonEmptyList` (@matthewleon)
+
+## [v4.2.0](https://github.com/purescript/purescript-lists/releases/tag/v4.2.0) - 2017-05-28
+
+- Reimplement `snoc` using `foldr` (@LiamGoodacre)
+- Use `~>` where appropriate (@matthewleon)
+- Add `repeat` and `iterate` for non-empty lazy lists (@matthewleon)
+
+## [v4.1.1](https://github.com/purescript/purescript-lists/releases/tag/v4.1.1) - 2017-05-09
+
+- Export `last` for `NonEmptyList`s
+
+## [v4.1.0](https://github.com/purescript/purescript-lists/releases/tag/v4.1.0) - 2017-04-27
+
+- Added `some` and `many` for `Data.List.Lazy` (@safareli)
+
+## [v4.0.1](https://github.com/purescript/purescript-lists/releases/tag/v4.0.1) - 2017-03-29
+
+- Restore compiler-optimized TCO operations
+
+## [v4.0.0](https://github.com/purescript/purescript-lists/releases/tag/v4.0.0) - 2017-03-26
+
+- Updated for PureScript 0.11.0
+- Swapped arguments of `mapWithIndex` for consistency with `purescript-arrays`
+- Various code style tweaks & warning fixes (@matthewleon)
+- Added `appendFoldable` for `NonEmptyList`s (@matthewleon)
+
+## [v3.4.0](https://github.com/purescript/purescript-lists/releases/tag/v3.4.0) - 2017-01-29
+
+Add `unsnoc` (@joshuahhh)
+
+## [v3.3.0](https://github.com/purescript/purescript-lists/releases/tag/v3.3.0) - 2016-12-27
+
+Use `~>` where appropriate (@mlang)
+
+## [v3.2.2](https://github.com/purescript/purescript-lists/releases/tag/v3.2.2) - 2016-12-24
+
+Stack safety for various functions (@sammthomson)
+
+## [v3.2.1](https://github.com/purescript/purescript-lists/releases/tag/v3.2.1) - 2016-11-14
+
+- Fixed shadowed name warnings
+
+## [v3.2.0](https://github.com/purescript/purescript-lists/releases/tag/v3.2.0) - 2016-11-02
+
+- Adds an implementation of `traverse` which does not grow the stack.
+
+## [v3.1.0](https://github.com/purescript/purescript-lists/releases/tag/v3.1.0) - 2016-10-28
+
+- Added `manyRec` and `someRec` - stack safe versions of `many` and `some` via `MonadRec`.
+
+## [v3.0.1](https://github.com/purescript/purescript-lists/releases/tag/v3.0.1) - 2016-10-28
+
+- Fix `purescript-generics` dependency to use `^` rather than `v`.
+
+## [v3.0.0](https://github.com/purescript/purescript-lists/releases/tag/v3.0.0) - 2016-10-16
+
+- The `group` functions now return `NonEmptyList` typed values rather than `NonEmpty List`
+- The lazy non-empty list is now lazier
+- The modules have been rearranged internally
+
+## [v2.1.0](https://github.com/purescript/purescript-lists/releases/tag/v2.1.0) - 2016-10-13
+
+- Added `Data.List.NonEmpty` (a `newtype` of `NonEmpty List a` with additional instances)
+- Added `Extend` instance for `List`
+
+## [v2.0.0](https://github.com/purescript/purescript-lists/releases/tag/v2.0.0) - 2016-10-09
+
+- Updated dependencies
+- The `group` functions now return `NonEmpty` groups
+- `show` now uses `:` in printing
+
+## [v1.0.1](https://github.com/purescript/purescript-lists/releases/tag/v1.0.1) - 2016-06-11
+
+- Added `Fail` instance for `Bind ZipList` (@joneshf)
+
+## [v1.0.0](https://github.com/purescript/purescript-lists/releases/tag/v1.0.0) - 2016-06-01
+
+This release is intended for the PureScript 0.9.1 compiler and newer.
+
+**Note**: The v1.0.0 tag is not meant to indicate the library is “finished”, the core libraries are all being bumped to this for the 0.9 compiler release so as to use semver more correctly.
+
+## [v1.0.0-rc.3](https://github.com/purescript/purescript-lists/releases/tag/v1.0.0-rc.3) - 2016-05-22
+
+- Added `Generic` instance for `List`
+
+## [v1.0.0-rc.2](https://github.com/purescript/purescript-lists/releases/tag/v1.0.0-rc.2) - 2016-05-20
+
+- Fixed errors and warnings
+
+## [v1.0.0-rc.1](https://github.com/purescript/purescript-lists/releases/tag/v1.0.0-rc.1) - 2016-03-25
+
+- Release candidate for the psc 0.8+ core libraries
+
+## [v0.7.10](https://github.com/purescript/purescript-lists/releases/tag/v0.7.10) - 2016-02-27
+
+Add `transpose` (@hdgarrood)
+
+## [v0.7.9](https://github.com/purescript/purescript-lists/releases/tag/v0.7.9) - 2016-01-11
+
+Add `fromFoldable`, `toUnfoldable` (@hdgarrood)
+
+## [v0.7.8](https://github.com/purescript/purescript-lists/releases/tag/v0.7.8) - 2016-01-06
+
+Make list `unfoldr` stack-safe (@rgrempel)
+
+## [v0.7.7](https://github.com/purescript/purescript-lists/releases/tag/v0.7.7) - 2015-11-02
+
+- Removed unused imports
+
+## [v0.7.6](https://github.com/purescript/purescript-lists/releases/tag/v0.7.6) - 2015-10-16
+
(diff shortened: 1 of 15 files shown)


src/Data/List.purs
```diff
@@ -632,18 +637,65 @@ tails list@(Cons _ tl)= list : tails tl
...
+  addIndexReverse
+  -- Sort by original values to cluster duplicates.
+  >>> sortBy (p `on` snd)
+  -- Removing neighboring duplicates.
+  >>> nubByAdjacentReverse (\a b -> (p `on` snd) a b == EQ)
+  -- Sort by index to recover original order.
+  -- Use `flip` to sort in reverse order in anticipation of final `mapReverse`.
+  >>> sortBy (flip compare `on` fst)
+  -- Discard indicies, just keep original values.
+  >>> mapReverse snd
```
milesfrain, line 670: I can't think of a much faster way to write this.
We could instead just use the more naive versions of the `O(n)` functions, considering that the sorts (`O(n log n)`) will take up the bulk of the time. But I figured it would be best to try for efficiency when possible.

src/Data/List.purs
```diff
@@ -763,3 +815,62 @@ transpose ((x : xs) : xss) =
...
+  where
+    go :: List a -> List a -> List a
+    -- empty output
+    go Nil (x : xs) = go (x : Nil) xs
+    -- checking for duplicates
+    go acc@(a : as) (x : xs)
+      | p a x = go (x : as) xs
+      | otherwise = go (x : acc) xs
+    -- empty input
+    go acc Nil = acc
```
milesfrain, line 876: These could just be helper functions for `nubBy`, rather than being exported. Not sure if anyone else would want to use these faster versions.

test/Test/Data/List.purs
```diff
@@ -398,6 +407,27 @@ testList = do
   log "append should be stack-safe"
   void $ pure $ xs <> xs
 
+  log "mapReverse should work the same as slower naive version"
+  --quickCheck mapReverseLaw
+
+  log "naiveAddIndexReverse should work the same as slower naive version"
+  --quickCheck naiveAddIndexReverse
```
milesfrain, line 414: I'm having trouble getting `quickcheck` to work as a bower dev dependency. I'll show you want I mean in the next commit. Maybe CI will show the same error.

src/Data/List.purs
```diff
@@ -632,18 +637,65 @@ tails list@(Cons _ tl)= list : tails tl
...
+-- |
+-- | Running time: `O(n log n)`
+nubBy :: forall a. (a -> a -> Ordering) -> List a -> List a
+nubBy p =
+  -- Add indices so we can recover original order after deduplicating.
+  addIndexReverse
+  -- Sort by original values to cluster duplicates.
+  >>> sortBy (p `on` snd)
+  -- Removing neighboring duplicates.
+  >>> nubByAdjacentReverse (\a b -> (p `on` snd) a b == EQ)
```
milesfrain, line 665: If we add an `Eq` constraint, this could be simplified to:
```purs
nubByAdjacentReverse (eq `on` snd)
```
That would probably also be faster, but seems to be against the spirit of allowing a custom ordering predicate.

src/Data/List.purs
```diff
@@ -632,18 +637,65 @@ tails list@(Cons _ tl)= list : tails tl
 --------------------------------------------------------------------------------
 
 -- | Remove duplicate elements from a list.
+-- | Keeps the first duplicates found and preserves ordering otherwise.
+-- |
+-- | ```purescript
+-- | nub 1:2:1:3:3:Nil = 1:2:3:Nil
```
milesfrain, line 643: Todo - change these examples to use `==`. Didn't catch this when converting from the Array docs. Will need to fix this in Array as well.

bower.json
```diff
@@ -37,6 +37,7 @@
     "purescript-assert": "master",
     "purescript-console": "master",
     "purescript-math": "master",
-    "purescript-minibench": "master"
+    "purescript-minibench": "master",
+    "purescript-quickcheck": "master"
```
milesfrain, line 41: After install `quickcheck` as a dev dependency, I can't build the project. Getting lots of "DuplicateModule" errors, such as:
```
Module Data.List has been defined multiple times
```

thomashoneyman: I'd assume that's because of a cyclical dependency:
https://github.com/purescript/purescript-quickcheck/blob/082f1f6681b6fc538e719622e35728d4de6d76e8/bower.json#L37

hdgarrood: Yep. I’m not aware of any good way around this unfortunately. I think if we put all of the core libraries into one big monorepo or even one big library we might be able to work around dependency issues like this, but neither of those are workable in the short term anyway. We will just have to do without for now. You could set up a separate project to perform these tests for peace of mind if that would help.

test/Test/Data/List.purs
```diff
@@ -398,6 +407,27 @@ testList = do
...
+
+  log "nubByAdjacentReverse should be correct"
+  assert $ nubByAdjacentReverse (on eq Array.length) ([1]:[2]:[3,4]:Nil) == [3,4]:[2]:Nil
+
+naiveMapReverse :: forall a b. (a -> b) -> List a -> List b
+naiveMapReverse f l = map f l # reverse
```
hdgarrood, line 420: For consistency, could you please use a right-to-left style here (and elsewhere)? So prefer `<<<` over `>>>`, and `$` over `#`.

test/Test/Data/List.purs
```diff
@@ -398,6 +407,27 @@ testList = do
   log "append should be stack-safe"
   void $ pure $ xs <> xs
 
+  log "mapReverse should work the same as slower naive version"
```
hdgarrood, line 410: I’d actually prefer to just test the functions we care about directly. If we have a few individual cases with inputs and expected outputs, plus one or two really big examples to test stack safety, that would be ideal I think. In my experience 100k elements is a good amount: enough to reliably cause stack overflows if there are any stack safety issues.

src/Data/List.purs
```diff
@@ -632,18 +637,65 @@ tails list@(Cons _ tl)= list : tails tl
...
+-- | ```
+-- |
+-- | Running time: `O(n log n)`
+nub :: forall a. Ord a => List a -> List a
+nub = nubBy compare
+
+-- | Remove duplicate elements from a list based on the provided ordering function.
+-- | Keeps the first duplicates found and preserves ordering otherwise.
```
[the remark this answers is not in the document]
milesfrain: Yes that sounds clearer.
I was originally worrying that "Removes all duplicates" could be interpreted as either:
- `nub [1, 2, 2] == [1]`
- `nub [1, 2, 2] == [1, 2]`

test/Test/Data/List.purs
milesfrain: Will do.
Note that the new `nubBy` uses lots of `>>>`, which I think is easier to read. Should that be changed to `<<<`?

src/Data/List.purs
milesfrain: Exports removed

bower.json
milesfrain: Quickcheck removed

test/Test/Data/List.purs
```diff
@@ -282,8 +284,15 @@ testList = do
...
-  let nubPred = \x y -> if odd x then false else x == y
-  assert $ nubBy nubPred (l [1, 2, 2, 3, 3, 4, 4, 1]) == l [1, 2, 3, 3, 4, 1]
+  let nubPred = compare `on` Array.length
+  assert $ nubBy nubPred (l [[1],[2],[3,4]]) == l [[1],[3,4]]
+
+  log "nubEq should remove duplicate elements from the list, keeping the first occurence"
+  assert $ nubEq (l [1, 2, 2, 3, 4, 1]) == l [1, 2, 3, 4]
+
+  log "nubByEq should remove duplicate items from the list using a supplied predicate"
+  let nubEqPred = \x y -> if odd x then false else x == y
```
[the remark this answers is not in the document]
milesfrain: Changed this to match the docs example, which satisfies laws

test/Test/Data/List.purs
milesfrain: Removed this test, since that function is no longer exported

milesfrain:
If this looks good for `Data.List`, then I'll apply the same changes to `Lazy` and `NonEmpty`.

I'm also thinking of refactoring the test code and extracting common material, so we don't have as much copy-paste going on between the three files.

I assume circular dependencies prevent us from upgrading the test framework to something that prints more info upon failures. I'm a fan of [`test-unit`](https://pursuit.purescript.org/packages/purescript-test-unit/14.0.0).

hdgarrood:
Yes, everything is very basic in the core libraries for this reason. test-unit is not an option both because of circular dependencies and because there is a policy that core libraries cannot depend on non-core libraries.

test/Test/Data/List.purs
hdgarrood: For that one I could go either way, so I wouldn't mind leaving it as-is. Wherever it's a simple enough case that either is equally clear, though, I think we should use right-to-left style.

hdgarrood: Yes please

hdgarrood:
This looks great so far :+1:

JordanMartinez:
This looks good to me!

JordanMartinez (approved):

src/Data/List.purs
```diff
@@ -794,3 +845,62 @@ transpose ((x : xs) : xss) =
...
+  where
+    go :: List a -> List a -> List a
+    -- empty output
+    go Nil (x : xs) = go (x : Nil) xs
+    -- checking for duplicates
+    go acc@(a : as) (x : xs)
+      | p a x = go (x : as) xs
+      | otherwise = go (x : acc) xs
+    -- empty input
+    go acc Nil = acc
```
thomashoneyman, line 906: Elsewhere the `where` definitions aren't indented

```suggestion
  go :: List a -> List a -> List a
  -- empty output
  go Nil (x : xs) = go (x : Nil) xs
  -- checking for duplicates
  go acc@(a : as) (x : xs)
    | p a x = go (x : as) xs
    | otherwise = go (x : acc) xs
  -- empty input
  go acc Nil = acc
```

thomashoneyman (approved):
Thanks for taking care of this! Minor formatting suggestions, but otherwise 👍 from me.

src/Data/List/Lazy.purs
```diff
@@ -70,8 +70,8 @@ module Data.List.Lazy
   , groupBy
   , partition
 
-  , nub
-  , nubBy
+  , nubEq
```
hdgarrood, line 73: It looks like this module no longer exports `nub` or `nubBy`. I think we should wait until we can provide an Ord-based nub before renaming the existing ones to `nubEq` and `nubByEq`.

test/Test/Data/List.purs
```diff
@@ -285,8 +287,15 @@ testList = do
   assert $ nub (l [1, 2, 2, 3, 4, 1]) == l [1, 2, 3, 4]
 
   log "nubBy should remove duplicate items from the list using a supplied predicate"
-  let nubPred = \x y -> if odd x then false else x == y
-  assert $ nubBy nubPred (l [1, 2, 2, 3, 3, 4, 4, 1]) == l [1, 2, 3, 3, 4, 1]
+  let nubPred = compare `on` Array.length
+  assert $ nubBy nubPred (l [[1],[2],[3,4]]) == l [[1],[3,4]]
+
+  log "nubEq should remove duplicate elements from the list, keeping the first occurence"
```
hdgarrood, line 293: typo: "occurrence"

src/Data/List/Lazy.purs
milesfrain: Does this mean we also need to add a `sort` for `Data.List.Lazy` too? That might be pretty involved. Would we want to match the Haskell version (see link in first post of this PR).

hdgarrood: Not necessarily - I would prefer to just leave the functions in this module alone then, if adding a sort implementation for use in an Ord-based nub would be involved. I just don’t want to remove nub from this module’s exports at any point; we shouldn’t require downstream users to rewrite existing uses of nub unnecessarily.

milesfrain: What about making Ord-based (according to the signature) `nub` and `nubBy`, which actually map to the slower Eq-based `nubEq` and `nubByEq` internally. We can add a comment in the docs that these versions are currently just as slow (or maybe even slightly slower) as the the Eq versions. Then later on, we can make the performance-boosting fix without affecting the API.

hdgarrood: I definitely don’t want to do that, it would be really misleading. I think I’d be very annoyed as a user if this function posed as O(n log n) by having an Ord signature, and then I used it and found that it was actually quadratic after it bit me in production.

milesfrain: I'd be happy to dive into writing the correct Ord-based `nub` and `nubBy`, but I don't yet have a good strategy in mind for that. Could you point me to some resources to help get me going, or maybe some pseudocode for how you would go about writing this?

hdgarrood: I don’t know of a resource and I’m not sure exactly how I’d go about it off the top of my head, sorry. I’d be happy with just leaving this module alone for now though. It does of course mean that we can’t provide an Ord-based nub until the next round of breaking changes, but if we do find ourselves wanting to get one in before then we can always add a `nubOrd` temporarily.

milesfrain: Just did some research and I think this is the ideal approach for sorting, and can be adapted for `nub`:
- Build a heap from the lazy list. This requires reading through the entire lazy list, and won't work on infinite lists.
- Lazily grab the head of the heap when consuming each element. This has a small performance boost versus traditional sorting if you only want a few elements.

But for our case, I think it's fine to forgoe the small perf boost for purely-lazy situations and just call `toFoldable`/`nub`(from list or array)/`fromFoldable`. This will still be O (n log n).

hdgarrood: Eh, I dunno, all of the options have downsides. What do other people think? I think my preference is
1. Leave this module alone for now, switch nub to use Ord in the next round of breaking changes
2. Implement sort and Ord-based nub for lazy lists
3. Provide a nub with Ord in the type signature which is really an Eq-based nub

hdgarrood: Oh to clarify, when I said “not necessarily,” that was only in response to “should I add sort,” not in response to “should I do it the way Haskell did.” I would suggest starting from Haskell’s implementation if you were going to look into implementing sort.

hdgarrood: Oh sorry, I only just saw your most recent comment. Going via strict lists or arrays sounds fine to me (I don’t have a preference for which).

hdgarrood: Ah, hm, actually maybe not. I think having sorting work by going via strict lists or arrays would be okay, but a nub function on lazy lists really ought to be able to handle eg `take 1 $ nub $ repeat “hi”`, I think. Especially if the current one does handle that (I think it does).

milesfrain: That's a good point, since this currently works:
```purs
take 2 $ nub $ cycle $ range 1 3
```

src/Data/List.purs
```diff
@@ -663,18 +663,64 @@ tails list@(Cons _ tl)= list : tails tl
...
+-- | nubByEq mod3eq 1:3:4:5:6:Nil == 1:3:5:Nil
+-- | ```
 -- |
 -- | Running time: `O(n^2)`
-nubBy :: forall a. (a -> a -> Boolean) -> List a -> List a
-nubBy _     Nil = Nil
-nubBy eq' (x : xs) = x : nubBy eq' (filter (\y -> not (eq' x y)) xs)
+nubByEq :: forall a. (a -> a -> Boolean) -> List a -> List a
+nubByEq _     Nil = Nil
+nubByEq eq' (x : xs) = x : nubByEq eq' (filter (\y -> not (eq' x y)) xs)
```
milesfrain, line 723: This was never stack-safe. Reported in #194

src/Data/List/Lazy.purs
```diff
@@ -590,21 +593,45 @@ partition f = foldr go {yes: nil, no: nil}
...
+nubBy :: forall a. (a -> a -> Ordering) -> List a -> List a
+nubBy p = go emptySet
+  where
+    go s (List l) = List (map (goStep s) l)
+    goStep _ Nil = Nil
+    goStep s (Cons a as) =
+      let { found, result: s' } = insertAndLookupBy p a s
+      in if found
+        then step (go s' as)
+        else Cons a (go s' as)
```
milesfrain, line 618: This doesn't look stack safe, but I don't know of a good way to fix it. The stack-safety strategy for strict `nubBy` involved reversing the final result, but that can't be done here with infinite lazy lists. Reported in #194

MonoidMusician: I think it is mostly stack-safe, because it's the caller's responsibility to force the thunks and most of the recursive calls are deferred. However, there could potentially be enough duplicate elements in a row to blow the stack (since we do recursively force thunks in that case).

CHANGELOG.md
```diff
@@ -5,8 +5,10 @@ Notable changes to this project are documented in this file. The format is based
 ## [Unreleased]
 
 Breaking changes:
+- Convert `nub`/`nubBy` to use ordering, rather than equality (#179)
 
 New features:
+- Add `nubEq`/`nubByEq` (#179)
```
milesfrain, line 11: Could have alternatively written this as:
> Breaking changes:
> - Rename `Eq`-based `nub`/`nubBy` to `nubEq`/`nubByEq` (#179)
>
> New features:
> - Add `Ord`-based `nub`/`nubBy` (#179)

test/Test/Data/List/Lazy.purs
```diff
@@ -332,8 +334,15 @@ testListLazy = do
   assert $ nub (l [1, 2, 2, 3, 4, 1]) == l [1, 2, 3, 4]
 
   log "nubBy should remove duplicate items from the list using a supplied predicate"
-  let nubPred = \x y -> if odd x then false else x == y
-  assert $ nubBy nubPred (l [1, 2, 2, 3, 3, 4, 4, 1]) == l [1, 2, 3, 3, 4, 1]
+  let nubPred = compare `on` Array.length
```
hdgarrood, line 337: Could you add a test which ensures that nub behaves sensibly on infinite lists please? Perhaps
```
log "nub should not consume more of the input list than necessary"
assert $ (take 3 $ nub $ cycle $ l [1,2,3]) == l [1,2,3]
```

CHANGELOG.md
```diff
@@ -5,22 +5,24 @@ Notable changes to this project are documented in this file. The format is based
...
-- Fix Lazy List docs where original list is returned instead of Nothing (#169)
-- Migrate to GitHub Actions (#177)
-- Change `foldM` type signature to more closely match `foldl` (#165)
-- Improve `foldr` performance on large lists (#180)
-- Generate changelog and add PR template (#187) 
+- Fixed Lazy List docs where original list is returned instead of Nothing (#169)
+- Migrated to GitHub Actions (#177)
+- Changed `foldM` type signature to more closely match `foldl` (#165)
+- Improved `foldr` performance on large lists (#180)
+- Generated changelog and add PR template (#187)
```
milesfrain, line 25: I believe consensus for changelog is to use past tense

test/Test/Data/List/Lazy.purs
milesfrain: Should this test be applied to all `nub*` functions, or just `nub`? I just added the single test.

hdgarrood: I think it’s worth doing it for nub and nubEq but we probably don’t need it for the *By variants, as if it works for nub it’ll probably work for nubBy too.

thomashoneyman:
I believe this PR can be merged, but I think it would be good for @hdgarrood and/or @MonoidMusician to have one more glance over it before we do.

MonoidMusician:
Looks good to me! Did a couple more tests in the repl just to verify

hdgarrood (approved):
Looks good to me :+1:

PR #180, opened 2020-12-19
Improve foldr performance ~4x on large lists

milesfrain:
Performance is improved by using an inlined reverse function, rather than leveraging `foldl`.

Both versions appear to be identical, so it's a bit of a bummer that the concise version is so much slower.

There's a moderate speedup for all list sizes, but it's most significant on large lists. Here are the benchmarking results:

Before:
```
foldr: list (0 elems)
mean   = 1.52 μs
stddev = 416.18 ns
min    = 1.22 μs
max    = 11.11 μs
---
foldr: list (1 elems)
mean   = 2.90 μs
stddev = 22.78 μs
min    = 977.00 ns
max    = 631.81 μs
---
foldr: list (10 elems)
mean   = 5.96 μs
stddev = 21.03 μs
min    = 2.00 μs
max    = 563.57 μs
---
foldr: list (100 elems)
mean   = 16.16 μs
stddev = 50.97 μs
min    = 10.32 μs
max    = 1.54 ms
---
foldr: list (1000 elems)
mean   = 126.53 μs
stddev = 77.85 μs
min    = 101.90 μs
max    = 1.40 ms
---
foldr: list (10000 elems)
mean   = 1.36 ms
stddev = 301.37 μs
min    = 1.17 ms
max    = 3.69 ms
---
foldr: list (100000 elems)
mean   = 22.66 ms
stddev = 3.63 ms
min    = 20.18 ms
max    = 43.71 ms
```
After:
```
foldr: list (0 elems)
mean   = 1.20 μs
stddev = 542.72 ns
min    = 772.00 ns
max    = 10.38 μs
---
foldr: list (1 elems)
mean   = 1.66 μs
stddev = 3.04 μs
min    = 806.00 ns
max    = 96.76 μs
---
foldr: list (10 elems)
mean   = 5.40 μs
stddev = 19.95 μs
min    = 1.05 μs
max    = 392.62 μs
---
foldr: list (100 elems)
mean   = 11.22 μs
stddev = 16.66 μs
min    = 4.67 μs
max    = 332.93 μs
---
foldr: list (1000 elems)
mean   = 53.27 μs
stddev = 40.41 μs
min    = 44.55 μs
max    = 529.10 μs
---
foldr: list (10000 elems)
mean   = 525.74 μs
stddev = 192.72 μs
min    = 450.25 μs
max    = 2.29 ms
---
foldr: list (100000 elems)
mean   = 4.66 ms
stddev = 2.23 ms
min    = 2.09 ms
max    = 12.26 ms
```

Also refactored the benchmarking code.

### DIFF
--- a/src/Data/List/Types.purs
+++ b/src/Data/List/Types.purs
@@ -103,7 +103,10 @@ instance functorWithIndexList :: FunctorWithIndex Int List where
 instance foldableList :: Foldable List where
   foldr f b = foldl (flip f) b <<< rev
     where
-    rev = foldl (flip Cons) Nil
+    rev = go Nil
+      where
+      go acc Nil = acc
+      go acc (x : xs) = go (x : acc) xs
   foldl f = go
     where
     go b = case _ of
--- a/bench/Bench/Data/List.purs
+++ b/bench/Bench/Data/List.purs
@@ -1,9 +1,10 @@
 module Bench.Data.List where
 
 import Prelude
-import Data.Maybe (fromMaybe)
 import Data.Foldable (maximum)
-import Data.List (List(..), take, range, foldr, length, (:))
+import Data.Int (pow)
+import Data.List (List(..), take, range, foldr, length)
+import Data.Maybe (fromMaybe)
 import Data.Traversable (traverse_)
 import Effect (Effect)
 import Effect.Console (log)
@@ -16,7 +17,7 @@ benchList = do
 
   where
 
-  listSizes = 0 : 1 : 1000 : 2000 : 5000 : 10000 : 100000 : Nil
+  listSizes = Cons 0 $ map (pow 10) $ range 0 5
   nats = range 0 $ (fromMaybe 0 $ maximum listSizes) - 1
   lists = map (\n -> take n nats) listSizes
 
--- a/bench/Bench/Data/List.purs
+++ b/bench/Bench/Data/List.purs
@@ -1,48 +1,31 @@
 module Bench.Data.List where
 
 import Prelude
+import Data.Maybe (fromMaybe)
+import Data.Foldable (maximum)
+import Data.List (List(..), take, range, foldr, length, (:))
+import Data.Traversable (traverse_)
 import Effect (Effect)
 import Effect.Console (log)
 import Performance.Minibench (bench)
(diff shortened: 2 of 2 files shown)


src/Data/List/Types.purs
```diff
@@ -103,7 +103,10 @@ instance functorWithIndexList :: FunctorWithIndex Int List where
 instance foldableList :: Foldable List where
   foldr f b = foldl (flip f) b <<< rev
     where
-    rev = foldl (flip Cons) Nil
+    rev = go Nil
+      where
+      go acc Nil = acc
+      go acc (x : xs) = go (x : acc) xs
```
milesfrain, line 109: This is the core change. Wondering if the `reverse` function should be moved from `Data.List` to this file to deduplicate code.

hdgarrood (approved):
Very nice! I’m not concerned about the duplication of `reverse`, since it’s so little code and it might look odd in the docs to say that `reverse` is re-exported from Data.List.Types whereas almost none of the other List functions are.

milesfrain:
This slower reverse pattern is duplicated in a few other places (just search for `foldl (flip`, for example:
https://github.com/purescript/purescript-lists/blob/62900a56f6864638c952575dfd211a1cc55be7da/src/Data/List/Types.purs#L135
https://github.com/purescript/purescript-lists/blob/62900a56f6864638c952575dfd211a1cc55be7da/src/Data/List/Types.purs#L141
https://github.com/purescript/purescript-lists/blob/62900a56f6864638c952575dfd211a1cc55be7da/src/Data/List/Types.purs#L145
https://github.com/purescript/purescript-lists/blob/62900a56f6864638c952575dfd211a1cc55be7da/src/Data/List/Types.purs#L153
https://github.com/purescript/purescript-lists/blob/62900a56f6864638c952575dfd211a1cc55be7da/src/Data/List/Types.purs#L260

So thinking these should all be replaced with a common `rev` helper function in this file (along with some comments about why it's written longform).

I'm also wondering if we should do more benchmarking in the browser (blocked by https://github.com/purescript/purescript-minibench/issues/16) with bundling optimizations and see if there's still a performance difference. Ideally, I'd like to keep the original concise code.

PR #181, opened 2020-12-20
Idiomatic Show instance for lazy lists

kl0tl:
This pull request changes the output of show for Data.List.Lazy.Types.List and Data.List.Lazy.Types.NonEmptyList from this:

```purs
> Data.List.Lazy.fromFoldable [0, 1]
fromStrict ((Cons 0 (Cons 1 Nil)))

> Data.List.Lazy.NonEmpty.fromFoldable [0, 1]
(Just (NonEmptyList (defer \_ -> (NonEmpty 0 fromStrict ((Cons 1 Nil))))))
```

to that:

```purs
> Data.List.Lazy.fromFoldable [0, 1]
(defer \_ -> 0 : (defer \_ -> 1 : (defer \_ -> Nil)))

> Data.List.Lazy.NonEmpty.fromFoldable [0, 1]
(Just (NonEmptyList (defer \_ -> (NonEmpty 0 (defer \_ -> 1 : (defer \_ -> Nil))))))
```

Alternatively, we could show `(fromFoldable […])` but this makes the output of show less similar:

```purs
> Data.List.Lazy.fromFoldable [0, 1]
(fromFoldable [1,0])

> Data.List.Lazy.NonEmpty.fromFoldable [0, 1]
(Just (NonEmptyList (defer \_ -> (NonEmpty 0 (fromFoldable [1])))))
```

Close https://github.com/purescript/purescript-lists/issues/122.

### DIFF
--- a/src/Data/List/Lazy/Types.purs
+++ b/src/Data/List/Lazy/Types.purs
@@ -34,6 +34,10 @@ newtype List a = List (Lazy (Step a))
 -- | `Cons` constructor).
 data Step a = Nil | Cons a (List a)
 
+instance showStep :: Show a => Show (Step a) where
+  show Nil = "Nil"
+  show (Cons x xs) = "(" <> show x <> " : " <> show xs <> ")"
+
 -- | Unwrap a lazy linked list
 step :: forall a. List a -> Step a
 step = force <<< unwrap
@@ -59,10 +63,12 @@ infixr 6 cons as :
 derive instance newtypeList :: Newtype (List a) _
 
 instance showList :: Show a => Show (List a) where
-  show xs = "fromStrict (" <> go (step xs) <> ")"
-    where
-    go Nil = "Nil"
-    go (Cons x xs') = "(Cons " <> show x <> " " <> go (step xs') <> ")"
+  show xs = "(fromFoldable ["
+         <> case step xs of
+              Nil -> ""
+              Cons x xs' ->
+                show x <> foldl (\shown x' -> shown <> "," <> show x') "" xs'
+         <> "])"
 
 instance eqList :: Eq a => Eq (List a) where
   eq = eq1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Breaking changes:
 - Renamed `scanrLazy` to `scanlLazy` and fixed parameter ordering (#161)
 - Renamed `group'` to `groupAll` (#182)
 - Changed `Alt ZipList` to satisfy distributivity (#150)
+- Updated the `Show` instances for (non empty) lazy lists (#181)
 
 New features:
 - Added `nubEq`/`nubByEq` (#179)


JordanMartinez (approved):

natefaubion:
While this is technically code that does indeed work, I'm not sure how useful it is as a Show instance since it's very much _not_ compact. It's very difficult to see what the values are in the midst of all the other noise (and parens, which will quickly get out of hand). I feel like it kind of misses the forest for the trees as far as helping someone see what's actually in the list.

milesfrain:
> I feel like it kind of misses the forest for the trees as far as helping someone see what's actually in the list.

Seems like this would be less of a dilemma if we adopted [debugged](https://github.com/hdgarrood/purescript-debugged).

src/Data/List/Lazy/Types.purs
```diff
@@ -34,6 +34,10 @@ newtype List a = List (Lazy (Step a))
 -- | `Cons` constructor).
 data Step a = Nil | Cons a (List a)
 
+instance showStep :: Show a => Show (Step a) where
+  show Nil = "Nil"
+  show (Cons x xs) = show x <> " : " <> show xs
```
natefaubion, line 39: Should this not be wrapped in parens?

natefaubion:
I personally think `fromFoldable` is the way to go for this.

hdgarrood:
I still want to adopt debugged in core, but in any case I think we should continue providing Show instances for the foreseeable future.

JordanMartinez:
Let's use `fromFoldable`. Who's going to do the work?

kl0tl:
I'll come back to this once I'm done with the compiler releases notes. I'll never get them done if I find other ways to spend m'y PureScript budget 😬

JordanMartinez:
Does this PR only need these two changes?
```purescript
-- adding parenthesis around `Step`'s `Cons` case
instance showStep :: Show a => Show (Step a) where
  show Nil = "Nil"
  show (Cons x xs) = "(" <> show x <> " : " <> show xs <> ")"

-- using the 'fromFoldable' approach:
instance showList :: Show a => Show (List a) where
  show ls = "fromFoldable [" <> print (step ls) <> "]"
    where
    print Nil = ""
    print (Cons x xs') = show x <> go (step xs')

    go Nil = ""
    go (Cons x xs') = ", " <> show x
```

JordanMartinez:
@kl0tl Mind if I push the changes I mentioned above here? If that's all we need to do, then we can get this PR merged.

kl0tl:
I pushed a patch based on your suggestion! Here’s the output of show for Data.List.Lazy.Types.List, Data.List.Lazy.Types.Step and Data.List.Lazy.Types.NonEmptyList:

```purs
> Data.List.Lazy.fromFoldable [0, 1, 2]
(fromFoldable [0,1,2])

> Data.List.Lazy.NonEmpty.fromFoldable [0, 1, 2]
(Just (NonEmptyList (defer \_ -> (NonEmpty 0 (fromFoldable [1,2])))))

> Data.List.Lazy.step (Data.List.Lazy.fromFoldable [0, 1, 2])
(0 : (fromFoldable [1,2]))
```

kl0tl:
I also opened https://github.com/purescript/purescript-ordered-collections/pull/46 for consistency, so that we display arrays instead of lists when showing (non empty) sets.

JordanMartinez (approved):

JordanMartinez:
Any other feedback before we merge this?

thomashoneyman (approved):

PR #188, opened 2021-01-11
Add Eq1 Ord1 to NonEmptyList LazyNonEmptyList

### LINKED ISSUE #186: Add Eq1 and Ord1 instances to NonEmptyList and LazyNonEmptyList
These instances exist for standard and Lazy lists. I assume we want them in the NonEmpty versions too for completeness.
Related to #183

milesfrain:
**Description of the change**

Adds `Eq1` and `Ord1` instances to `NonEmptyList` and `LazyNonEmptyList`
Fixes #186

---

**Checklist:**

- [x] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [x] Linked any existing issues or proposals that this pull request should close
- [x] Updated or added relevant documentation

### DIFF
--- a/src/Data/List/Lazy/Types.purs
+++ b/src/Data/List/Lazy/Types.purs
@@ -211,8 +211,11 @@ derive instance newtypeNonEmptyList :: Newtype (NonEmptyList a) _
 derive newtype instance eqNonEmptyList :: Eq a => Eq (NonEmptyList a)
 derive newtype instance ordNonEmptyList :: Ord a => Ord (NonEmptyList a)
 
-derive newtype instance eq1NonEmptyList :: Eq1 NonEmptyList
-derive newtype instance ord1NonEmptyList :: Ord1 NonEmptyList
+instance eq1NonEmptyList :: Eq1 NonEmptyList where
+  eq1 (NonEmptyList lhs) (NonEmptyList rhs) = eq1 lhs rhs
+
+instance ord1NonEmptyList :: Ord1 NonEmptyList where
+  compare1 (NonEmptyList lhs) (NonEmptyList rhs) = compare1 lhs rhs
 
 instance showNonEmptyList :: Show a => Show (NonEmptyList a) where
   show (NonEmptyList nel) = "(NonEmptyList " <> show nel <> ")"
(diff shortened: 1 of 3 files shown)


milesfrain:
CI shows that the Lazy version of this change fails to compile, but the non-Lazy version works fine. Don't know why.

src/Data/List/Lazy/Types.purs
```diff
@@ -211,6 +211,12 @@ derive instance newtypeNonEmptyList :: Newtype (NonEmptyList a) _
 derive newtype instance eqNonEmptyList :: Eq a => Eq (NonEmptyList a)
 derive newtype instance ordNonEmptyList :: Ord a => Ord (NonEmptyList a)
 
+instance eq1NonEmptyList :: Eq1 NonEmptyList where
+  eq1 (NonEmptyList lhs) (NonEmptyList rhs) = eq1 lhs rhs
```
milesfrain, line 215: Was hoping this shorthand would work, but it does not:
```purs
eq1 = over NonEmptyList eq1
```

thomashoneyman (approved):

src/Data/List/Lazy/Types.purs
thomashoneyman: I've seen a number of cases where omitting arguments in an instance doesn't work. But I think it's arguably better to include the arguments explicitly in any case.

milesfrain:
Should these changes be tested, and if so, I guess I should also add the same tests to the empty-able lists too?
Also, holding off on editing the changelogs for each of these PR, since that's going to cause a merge conflict after each PR is merged. Will do that right before 0.14

hdgarrood (approved):
I don’t think these need tests - the type system helps us enough here (via parametricity) that I think we can be confident these are doing the right thing without needing additional tests.

hdgarrood:
Actually let’s include the changelog updated in each PR please; the merge conflicts that would be introduced will all be very easy to fix.

hdgarrood:
Thank you!

src/Data/List/Lazy/Types.purs
milesfrain: I'd argue that the shorthand version is better, since it protects you from making a mistake like this:
```purs
eq1 (NonEmptyList rhs) (NonEmptyList lhs) = eq1 lhs rhs
```

PR #190, opened 2021-01-13
Infix operator (:||) for NonEmptyList cons'

milesfrain:
**Description of the change**

Motivation for this change is to give users a more ergonomic alternative to:
https://github.com/purescript/purescript-lists/blob/6d8e30eeb9714bc422b6fb75ad1b0f43bef59703/src/Data/List.purs#L619-L620

We can currently do a little better with NonEmpty's `:|` operator:
```purs
(NonEmptyList (1 :| 3 : Nil)) : (NonEmptyList (2 :| Nil)) : (NonEmptyList (4 :| Nil)) : (NonEmptyList (3 :| (3 : Nil))) : Nil
```

But it's even nicer with this change, which introduces the `:||` operator for `cons'`.
```purs
(1 :|| 3 : Nil) : (2 :|| Nil) : (4 :|| Nil) : (3 :|| 3 : Nil) : Nil
```

I'd ideally like to reuse the `:|` operator, but that leads to complications since we then can't use that operator for pattern matching anymore (it's equivalent to `NonEmptyList (NonEmpty x xs)`. The new version is no longer a simple data constructor of 2 arguments. We'd likely need to enable [Pattern Synonyms](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pattern_synonyms.html) for that to work.

---

**Checklist:**

- [ ] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [ ] Linked any existing issues or proposals that this pull request should close
- [ ] Added a test for the contribution (if applicable)

### DIFF
--- a/src/Data/List/NonEmpty.purs
+++ b/src/Data/List/NonEmpty.purs
@@ -7,7 +7,7 @@ module Data.List.NonEmpty
   , singleton
   , length
   , cons
-  , cons'
+  , (:||), cons'
   , snoc
   , snoc'
   , head
@@ -93,7 +93,7 @@ wrappedOperation
   -> NonEmptyList b
 wrappedOperation name f (NonEmptyList (x :| xs)) =
   case f (x : xs) of
-    x' : xs' -> NonEmptyList (x' :| xs')
+    x' : xs' -> x' :|| xs'
     L.Nil -> unsafeCrashWith ("Impossible: empty list in NonEmptyList " <> name)
 
 -- | Like `wrappedOperation`, but for functions that operate on 2 lists.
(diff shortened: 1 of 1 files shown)


src/Data/List/NonEmpty.purs
```diff
@@ -132,16 +133,18 @@ singleton :: forall a. a -> NonEmptyList a
...
 
 cons' :: forall a. a -> L.List a -> NonEmptyList a
-cons' x xs = NonEmptyList (x :| xs)
+cons' x xs = NonEmptyList (NonEmpty x xs)
+
+infixr 5 cons' as :||
 
 snoc :: forall a. NonEmptyList a -> a -> NonEmptyList a
-snoc (NonEmptyList (x :| xs)) y = NonEmptyList (x :| L.snoc xs y)
+snoc (NonEmptyList (x :| xs)) y = x :|| L.snoc xs y
```
milesfrain, line 144: If we could get pattern matching to work, and reuse `:|`, we could rewrite this (and friends) as:
```purs
snoc (x :| xs) y = x :| L.snoc xs y
```

hdgarrood:
Adding this would be non-breaking so I'd suggest leaving this until after v0.14.0.

JordanMartinez:
@hdgarrood Why wait to merge this if it builds and is otherwise acceptable?

hdgarrood:
Because we should be focusing on getting v0.14.0 out - it's been in the pipeline for far too long already.

Masynchin:
What is the current status of this proposal? It is planned to be merged?

Issue #184, opened 2020-12-23
Improve consistency with other libraries


milesfrain:
There's a lot in common among containers such as `List`, `Array`, `Set`, but their APIs drift out of sync. What's a good strategy to keep everything up to date?

We could continue to apply techniques such as #183 across these libraries, but that requires constant manual intervention.

**Would it be appropriate to use more type classes to enforce common interfaces?** Or should we avoid those, as suggested in https://harry.garrood.me/blog/down-with-show-part-1/ ? I still don't have a good understanding of how using more typeclasses leads to trouble.

For example we could have a `SetLike` type class that provides `intersection`, `union`, `difference`, etc. that `List`, `Array`, `Set` would all have an instance of. A similar strategy could have also helped us avoid https://github.com/purescript/purescript-lists/issues/178 and needing to remember to address https://github.com/purescript/purescript-arrays/issues/192. Interfaces could also cut down on the burden of maintaining duplicated documentation for all of these duplicated functions. 

Another minor bit of messiness is that we currently have both [`intersect`](https://pursuit.purescript.org/search?q=intersect) and [`intersection`](https://pursuit.purescript.org/search?q=intersection), depending on the container. Of course this can be fixed without resorting to typeclasses, but could also have been prevented by embracing more typeclasses.

-------------

Also, what determines whether a typeclass instance is defined in the type or the class? It seems somewhat arbitrary. For example:
- `Semigroup` defines the instances for `Array` and `Void`
- `Void` defines the instance for `Show`
- `Show` defines the instance for `Array`

Things would be a bit more organized if the policy in core was to always define instances in the type, which would look like the following:
- `Void` defines the instance for `Show` and `Semigroup`
- `Array` defines instances for `Show` and `Semigroup`

JordanMartinez:
> What's a good strategy to keep everything up to date?

I think adding a PR template that has a checkbox saying, "If you are adding a new API in library X, you must also open issues to add the same API in libraries Y and Z.

> I still don't have a good understanding of how using more typeclasses leads to trouble.

Here's my (potentially wrong) understanding. Type class constraints translate to type class dictionaries, which are just extra values passed into a given function. As a result, performance tends to suffer a bit because an extra closure is allocated. 

> Also, what determines whether a typeclass instance is defined in the type or the class?

I don't think there is a policy here. If a type class and a type appear in the same package, so that we could put the instance in either, I'm not sure which is preferable. 
The first issue is avoiding cyclical dependencies between modules. 
[...]

PR #199, opened 2021-04-22
Deduplicate testing code

### LINKED ISSUE #183: Patch API differences
It looks like there are a few missing `functions` and `instances` across the different list modules in this library. To tackle this more systematically, I found the contents of each of the cells of a 4-way Venn diagram created by comparing the following modules:
- Basic (List)
- Lazy[List]
- NonEmpty[List]
- LazyNonEmpty[List]

I've only looked over the `instances` so far. **These appear to be the gaps**:
- `Lazy` missing from `LazyNonEmptyList`
- `Eq1`, `Ord1` missing from `NonEmptyList`, `LazyNonEmptyList`
- `Foldable1`, `Traversable1` missing from `LazyNonEmptyList`

Here are some notes on the remaining differences, which appear to be fine:
- Missing from `*NonEmptyList`
  ```
  Monoid - Needs empty
  Plus - Needs empty
  Alternative - Needs Plus
  MonadZero - Needs Alternative
  MonadPlus - Needs MonadZero
  Unfoldable - Only if you can produce an empty container
  ```
- `Comonad` is only in `*NonEmptyList`, since you can't `extract` something from an empty list.
- Only non-basic lists need the `Newtype` convenience wrapper.

I still need to look through the `functions`. Feel free to help me flag or dismiss any of those differences.

----

Here's a brief explanation of how to read an entry in the listing of cells. For example in this entry:
```
-----
LazyNonEmpty NonEmpty
-----
Comonad
```
It means that `Comonad` appears only in `LazyNonEmpty` and `NonEmpty`, but not in any other modules. More specifically, `Comonad` is the output of:
```
intersections [LazyNonEmpty, NonEmpty] `difference` unions [Basic, NonEmpty]
```

This diagram might help with visualizing the "cells":

Here's the full listing (empty cells omitted).

**Instances:**
```
-----
Lazy LazyNonEmpty Basic NonEmpty
-----
Alt
Applicative
Apply
Bind
Eq
Extend
Foldable
FoldableWithIndex
Functor
FunctorWithIndex
Monad
Ord
Semigroup
Show
Traversable
TraversableWithIndex
Unfoldable1

-----
Lazy LazyNonEmpty NonEmpty
-----
Newtype

-----
LazyNonEmpty NonEmpty
-----
Comonad

-----
NonEmpty
-----
Foldable1
Traversable1

-----
Lazy Basic
-----
Alternative
Eq1
MonadPlus
MonadZero
Monoid
Ord1
Plus
Unfoldable

-----
Lazy
-----
Lazy
```

**Functions:**
```
-----
Lazy LazyNonEmpty Basic NonEmpty
-----
concatMap
fromFoldable
head
init
last
length
singleton
tail
toUnfoldable
uncons

-----
Lazy Basic NonEmpty
-----
(!!), index
catMaybes
concat
drop
dropWhile
elemIndex
elemLastIndex
filter
filterM
findIndex
findLastIndex
foldM
group
groupBy
insertAt
intersect
intersectBy
mapMaybe
modifyAt
nub
nubBy
partition
reverse
snoc
span
take
takeWhile
union
unionBy
unzip
updateAt
zip
zipWith
zipWithA

-----
Basic NonEmpty
-----
group'
mapWithIndex
sort
sortBy
unsnoc

-----
LazyNonEmpty NonEmpty
-----
appendFoldable
fromList
toList

-----
NonEmpty
-----
cons
cons'
snoc'

-----
Lazy Basic
-----
(..), range
(\\), difference
Pattern(..)
alterAt
delete
deleteAt
deleteBy
insert
insertBy
many
null
slice
some
stripPrefix
transpose

-----
Basic
-----
dropEnd
manyRec
someRec
takeEnd

-----
Lazy LazyNonEmpty
-----
iterate
repeat

-----
Lazy
-----
cycle
foldrLazy
replicate
replicateM
scanrLazy
```

milesfrain:
**Description of the change**

This PR demonstrates another testing strategy that accomplishes the following:
* Encourages API consistency
  * It lets the compiler help us fix #183 and prevent future API gaps.
* Reuses tests
  * Common tests were previously duplicated 4+ times for the different list types.
  * This strategy could be extended to allow reuse across a variety of container/collection types.
    * It looks like many of the tests in this lists library were initially copied over from the arrays library, so it would be great if we could let these be reused when possible.

This new testing strategy involves creating testing-specific typeclasses that cluster common APIs. This lets the compiler enforce common interfaces during library development, but doesn't expose this misuse of typeclasses to library consumers (see #184 for more discussion on typeclasses as interfaces).

These changes additive at the moment, but we could also delete any duplicated tests as they are migrated to these common tests. The goal is to migrate all preexisting tests.

---

**Checklist:**

- [ ] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [ ] Linked any existing issues or proposals that this pull request should close

### DIFF
--- a/test/Test/Common.purs
+++ b/test/Test/Common.purs
@@ -1,0 +1,547 @@
+module Test.Common where
+
+import Prelude
+
+import Control.Alt (class Alt, (<|>))
+import Control.Extend (class Extend, (<<=))
+import Data.Array as Array
+import Data.Eq (class Eq1)
+import Data.Foldable (class Foldable, foldMap, foldl, sum)
+import Data.FoldableWithIndex (class FoldableWithIndex, foldMapWithIndex, foldlWithIndex, foldrWithIndex)
+import Data.Function (on)
+import Data.FunctorWithIndex (class FunctorWithIndex, mapWithIndex)
+import Data.Int (odd)
+import Data.Maybe (Maybe(..), fromJust)
+import Data.Monoid.Additive (Additive(..))
+import Data.Ord (class Ord1)
+import Data.Traversable (class Traversable, traverse)
+import Data.TraversableWithIndex (class TraversableWithIndex, traverseWithIndex)
+import Data.Tuple (Tuple(..))
+import Data.Unfoldable (class Unfoldable, replicate, replicateA, unfoldr)
+import Data.Unfoldable1 (class Unfoldable1, unfoldr1)
+import Effect (Effect)
+import Effect.Console (log)
+import Partial.Unsafe (unsafePartial)
+import Test.Assert (assert)
+
+import Data.List as L
+import Data.List.NonEmpty as NEL
+import Data.List.Lazy as LL
+import Data.List.Lazy.NonEmpty as LNEL
+
+{-
+This is temporarily being used during development.
+It allows testing while still patching the API.
+This is passed as an additional argument to testCommon
+to indicate which container type is being tested, and
+lets us skip gaps that are currently implemented by `unsafeCrashWith`:
+
+Once fully supported by all containers, can replace with original assert.
+-}
+data SkipBroken
+  = SkipBrokenStrictCanEmpty
+  | SkipBrokenStrictNonEmpty
+  | SkipBrokenLazyCanEmpty
+  | SkipBrokenLazyNonEmpty
+  | RunAll
+
+derive instance eqSkipBroken :: Eq SkipBroken
+
+assertSkipHelper :: SkipBroken -> Array SkipBroken -> (_ -> Boolean) -> Effect Unit
+assertSkipHelper skip arr f =
+  case Array.elem skip arr of
+    true -> log "...skipped"
+    false -> assert $ f unit
+
+printContainerType :: String -> Effect Unit
+printContainerType str = do
+  log "--------------------------------"
+  log str
+  log "--------------------------------"
+
+printTestType :: String -> Effect Unit
+printTestType str = do
+  log $ "---- " <> str <> " Tests ----"
+
+class (
+  Alt c
+  , Applicative c
+  , Apply c
+  , Bind c
+  , Eq (c Int)
+  --, Eq1 c -- missing from NonEmptyList, LazyNonEmptyList
+  , Extend c
+  , Foldable c
+  , FoldableWithIndex Int c
+  , Functor c
+  , FunctorWithIndex Int c
+  , Monad c
+  , Ord (c Int)
+  --, Ord1 c -- missing from NonEmptyList, LazyNonEmptyList
+  , Semigroup (c Int)
+  , Show (c Int)
+  , Traversable c
+  , TraversableWithIndex Int c
+  , Unfoldable1 c
+) <= Common c where
+  concat :: forall a. c (c a) -> c a
+  concatMap :: forall a. forall b. (a -> c b) -> c a -> c b
+  -- Should basic list have a cons function wrapping the Cons constructor?
+  cons :: forall a. a -> c a -> c a
+  elemIndex :: forall a. Eq a => a -> c a -> Maybe Int
+  elemLastIndex :: forall a. Eq a => a -> c a -> Maybe Int
+  findIndex :: forall a. (a -> Boolean) -> c a -> Maybe Int
+  findLastIndex :: forall a. (a -> Boolean) -> c a -> Maybe Int
+  foldM :: forall m a b. Monad m => (b -> a -> m b) -> b -> c a -> m b
+  index :: forall a. c a -> Int -> Maybe a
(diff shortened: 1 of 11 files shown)


test/Test/Common.purs
```diff
@@ -0,0 +1,547 @@
...
+  snoc :: forall a. c a -> a -> c a
+  toUnfoldable :: forall f a. Unfoldable f => c a -> f a
+  union :: forall a. Eq a => c a -> c a -> c a
+  unionBy :: forall a. (a -> a -> Boolean) -> c a -> c a -> c a
+  -- Types don't have to be all a
+  -- Todo - double check this requirement
+  unzip :: forall a b. c (Tuple a b) -> Tuple (c a) (c b)
+  zip :: forall a b. c a -> c b -> c (Tuple a b)
+  zipWith :: forall a b d. (a -> b -> d) -> c a -> c b -> c d
+  zipWithA :: forall a b d m. Applicative m => (a -> b -> m d) -> c a -> c b -> m (c d)
```
milesfrain, line 113: Here's the example of APIs that are available to all list types. There should probably be more added here.

thomashoneyman: Are we certain that all "list" types can (or should) implement the same APIs? It seems that at least some operations would be useful for some list types and not others -- for example, I'd imagine some operations don't make sense for a lazy list.

thomashoneyman: I don't know the best place to have this discussion -- perhaps in the related "Patch API differences" conversation -- but my gut sense is that abstracting over list types in this way may have undesirable consequences for a) performance and b) error messages, and that we need to weigh that against the advantages of a unified interface.

milesfrain: > Are we certain that all "list" types can (or should) implement the same APIs? It seems that at least some operations would be useful for some list types and not others -- for example, I'd imagine some operations don't make sense for a lazy list.

You are correct. I just linked to the most "universally common" snippet, but there are testing clusters in this PR that target just `Lazy` vs `"Strict"` lists or `"Emptiable"` vs `"NonEmpty"` lists. I think looking at the `UpdatedTests.purs` file in this PR is a good place to see how everything fits together.

> my gut sense is that abstracting over list types in this way may have undesirable consequences for a) performance and b) error messages, and that we need to weigh that against the advantages of a unified interface.

This abstraction is just for managing and running tests, so shouldn't affect end users of the library.
But regarding error messages during testing, I'd love to be able to use [`test-unit`](https://pursuit.purescript.org/packages/purescript-test-unit) (or similar) instead of `assert` so we can more easily see what's causing the failure.

thomashoneyman: > I'd love to be able to use test-unit (or similar) instead of assert so we can more easily see what's causing the failure.

Core libraries only depend on other core libraries, so improvements to the testing situation would have to be added to `assert`; we won't be able to depend on other testing libraries, unfortunately.

milesfrain: > I'd still like to hear your thoughts about what we should do when a library needs to diverge from the standard API -- how would you envision us handling that?

Currently, this is just a common API for this lists library, but to answer two related questions:

What if one of the sub-list types (e.g. `LazyNonEmpty`) needs to diverge from the common List API?
- In that case, the offending function is no longer common across all list types, and can be broken-out into whatever cluster that function is still common among.

What if we use a common API among multiple containers (e.g. Arrays and Lists) and some List functions need to diverge.
- Similar answer to the above. If those functions aren't common across all containers, then they can't be part of the common-container API, but they can be demoted to part of the common-list API.

milesfrain:
Having issues reproducing the CI failure locally. Don't know why. I believe versions are the same:
```bash
➜  lists git:(dedup-tests) ✗ pulp --version        
Pulp version 15.0.0
purs version 0.14.1 using /home/miles/.nvm/versions/node/v14.9.0/bin/purs
➜  lists git:(dedup-tests) ✗ pulp build -- --censor-lib --strict --censor-codes='UserDefinedWarning'
* Building project in /home/miles/projects/purescript/lists
Invalid option `--censor-lib'

Usage: purs COMMAND
  The PureScript compiler and tools
* ERROR: Subcommand terminated with exit code 1
```

thomashoneyman:
Looks like you don’t have psa installed — you can either install it or omit the flags passed to pulp. The error you’re seeing is because of a compiler warning introduced in 0.14.1 that warns on unused names.

test/Test/Common.purs
milesfrain: > I'd love to be able to use [`test-unit`](https://pursuit.purescript.org/packages/purescript-test-unit) (or similar) instead of `assert` so we can more easily see what's causing the failure.

I forgot we had `assertEqual`, which works fine for displaying diffs.

PR #191, opened 2021-01-13
Revise groupAllBy to just use an Ordering function

milesfrain:
**Description of the change**

Fixes #189

Also modified the tests and examples to demonstrate which grouping functions involve sorting.

One of the commits reverts demoing the convenience of the `:||` infix operator proposed in #190. We can redo those changes if that operator is accepted.

---

**Checklist:**

- [x] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [x] Linked any existing issues or proposals that this pull request should close
- [x] Updated or added relevant documentation
- [x] Added a test for the contribution (if applicable)

PR #200, opened 2021-05-24
Deduplicate testing code - with records

milesfrain:
A version of #199 using records instead of typeclasses

WIP - A bit messy at the moment.


test/Test/AllTests.purs
```diff
@@ -0,0 +1,1099 @@
+module Test.AllTests where
```
milesfrain, line 1: All tests are merged into this file.

test/Test/API.purs
```diff
@@ -0,0 +1,197 @@
+module Test.API where
```
milesfrain, line 1: This file outlines the API for the list varieties

test/Test/UpdatedTests.purs
```diff
@@ -0,0 +1,105 @@
+module Test.UpdatedTests(updatedTests) where
```
milesfrain, line 1: This is where we call the common tests for each list variety

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ Notable changes to this project are documented in this file. The format is based
 ## [Unreleased]
 
 Breaking changes:
+- Rename `scanrLazy` to `scanlLazy` and fix parameter ordering (#161)
+- Rename `group'` to `groupAll` (#182)
+- Change `Alt ZipList` to satisfy distributivity (#150)
```
JordanMartinez, line 11: Is there a way to implement this PR without the above breaking changes? The changes could still be made in a separate PR that we could wait to merge until another breaking PS release is made.

src/Data/List.purs
```diff
@@ -180,6 +200,15 @@ manyRec p = tailRecM go Nil
     aa <- (Loop <$> p) <|> pure (Done unit)
     pure $ bimap (_ : acc) (\_ -> reverse acc) aa
 
+-- Questionable whether this should be specialized
+-- -- | Create a list containing a value repeated n times
+-- replicate :: forall a. Int -> a -> List a
+-- replicate num x = go num Nil
+--   where
+--   go n xs | n < 1 = xs
+--           | otherwise = go (n - 1) (x : xs)
```
JordanMartinez, line 209: I'm wondering whether this should be added. Does anyone use this? Would anyone use this?

src/Data/List/NonEmpty.purs
```diff
@@ -133,6 +188,14 @@ toList (NonEmptyList (x :| xs)) = x : xs
 singleton :: forall a. a -> NonEmptyList a
 singleton = NonEmptyList <<< NE.singleton
 
+-- | An infix synonym for `range`.
+infix 8 range as ..
+
+-- | Create a list containing a range of integers, including both endpoints.
+-- Todo, rewrite this without unsafe workaround (if necessary)
+range :: Int -> Int -> NonEmptyList Int
+range start end = unsafePartial fromJust $ fromList $ L.range start end
```
JordanMartinez, line 197: Perhaps this should be rewritten so that the `start` arg is always included in the range and guarantees the NonEmpty part of the returned value, so we don't need to wrap it in a `Maybe`?

test/Test/API.purs
```diff
@@ -0,0 +1,197 @@
...
+  -- , replicate :: forall a. Int -> a -> c a
+  , some :: forall f a. Alternative f => Lazy (f (c a)) => f a -> f (c a)
+  , someRec :: forall f a. MonadRec f => Alternative f => f a -> f (c a)
+  , sort :: forall a. Ord a => c a -> c a
+  , sortBy :: forall a. (a -> a -> Ordering) -> c a -> c a
+  , transpose :: forall a. c (c a) -> c (c a)
+  }
+
+type CommonDiffEmptiability c cInverse canEmpty nonEmpty cPattern =
+  { makeCollection :: forall f a. Foldable f => f a -> c a
```
JordanMartinez, line 56: To reduce verbosity and help show similarities (though I don't think this would help readability), should items like `makeCollection` be stored as a `Row` that gets added into each differing item via row polymorphism?

CHANGELOG.md
hdgarrood: These changes have already been made and released; I think this is just a bad merge.

milesfrain: I'm planning on merging this after #191 goes though, so I incorporated changes from that PR here too.
But yes, this is a pretty sloppy PR. Just wanted to upload the WIP to share some context during yesterday's meetup.

src/Data/List.purs
milesfrain: I'm wondering the same. Current plan is to just provide a specialized `replicate` for `List.Lazy`, and otherwise use `Unfoldable`'s `replicate` here. So I'd remove this commented block during clean-up.
The full API proposal can be found in https://github.com/purescript/purescript-lists/issues/183#issuecomment-825460643. Some related questions about what to include or not can be found at the end of that comment.

test/Test/API.purs
milesfrain: That could be a nice adjustment, but not sure the deduplication would be enough to justify the readability tradeoff and Row + Record overhead. I might be misimagining what changes are involved with this proposal though, so would be interested in seeing it implemented.

Another option could be to do something like this:
https://github.com/milesfrain/purescript-lists/pull/4/files

src/Data/List/NonEmpty.purs
JordanMartinez: I think the second one looks good.

test/Test/API.purs
JordanMartinez: Yeah... I'm not sure it's worth it either due to the readability tradeoff.

PR #220, opened 2025-02-27
Make `Data.List.NonEmpty.toUnfoldable` and `Data.List.Lazy.NonEmpty.toUnfoldable` work with `Unfoldable1`

UnrelatedString:
**Description of the change**

Modified `toUnfoldable` in `Data.List.NonEmpty` and `Data.List.Lazy.NonEmpty` to have the signature `toUnfoldable :: forall f. Unfoldable1 f => NonEmptyList ~> f`, and reimplemented it accordingly. Additionally added tests covering all four `toUnfoldable`s, since apparently there were no such tests in the existing codebase.

The motivation for this PR is that some users may hypothetically wish to convert known non-empty lists into other structures which cannot be empty. This can be addressed unobtrusively because `Unfoldable1` is a superclass of `Unfoldable`; the old type signature is completely compatible with the new implementation. For lack of a real use case, I was inspired to contribute this change when developing an even more superfluous package of my own which pervasively concerns itself with laziness in unfoldables.

Did not bump the version--although I don't think I could have made any breaking changes, I still changed the types of two functions in the public API (and the only relevant tests are ones I wrote), and either way I assume versioning decisions are best left to package maintainers.

---

**Checklist:**

- [x] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [x] Linked any existing issues or proposals that this pull request should close

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Breaking changes:
 
 New features:
 
+- Change `NonEmpty.toUnfoldable` to produce any `Unfoldable1` (#218 by @UnrelatedString)
+
 Bugfixes:
 
 Other improvements:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Breaking changes:
 
 New features:
 
-- Change `NonEmpty.toUnfoldable` to produce any `Unfoldable1` (#218 by @UnrelatedString)
+- Change `NonEmpty.toUnfoldable` to produce any `Unfoldable1` (#220 by @UnrelatedString)
 
 Bugfixes:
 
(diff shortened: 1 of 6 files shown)


UnrelatedString:
Have not added or updated any documentation because there is no documentation in either of the affected modules. I'm tempted to copy all of the documentation over from vanilla `Data.List`, but I assume this has not already been done to avoid future effort spent on mirroring any future documentation changes?

UnrelatedString:
Instead of replacing the existing `toUnfoldable`, this could also provide a separate `toUnfoldable1` like in Data.Array.NonEmpty. Seems a bit redundant to me, but I suppose it would also perform a bit better for `Unfoldable` containers if they implement `Unfoldable1` with an adapter to `unfoldr`.

srghma:
yes, very important, allows to `NonEmptyList -> NonEmptyArray`. Currently there is no way.

